PR #10, opened 2023-01-17
fix(ci): also push major floating tag and pre release

LucaGuerra:
Signed-off-by: Luca Guerra <luca@guerra.sh>

This adds, e.g. the tag `0` when we push something like `0.1`, also ensures that pre releases are also pushed, even if not latest.


loresuso (approved):
LGTM!

leogr (approved):

PR #16, opened 2023-01-20
new(build/registry): add command to upload rulesfiles to S3 and CI/CD trigger

LucaGuerra:
Signed-off-by: Luca Guerra <luca@guerra.sh>

As per title. This uploads files to a `rules` directory in the Falco bucket (the one displayed on download.falco.org)

NOTE FOR THE REVIEWER: see inline comments


build/registry/s3.go
```diff
@@ -0,0 +1,33 @@
...
+	upFile.Read(fileBuffer)
+
+	_, err = s3.New(session).PutObject(&s3.PutObjectInput{
+		Bucket:               aws.String(bucket),
+		Key:                  aws.String(key),
+		Body:                 bytes.NewReader(fileBuffer),
+		ContentLength:        aws.Int64(fileSize),
+		ContentDisposition:   aws.String("attachment"),
+		ServerSideEncryption: aws.String("AES256"),
+	})
```
LucaGuerra, line 31: Note for the reviewer: do you know if we need to set additional permissions or ACL when uploading this in the Falco bucket?

.github/workflows/release.yaml
```diff
@@ -33,3 +39,16 @@ jobs:
           REGISTRY_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GITHUB_REPO_URL: ${{ github.server_url }}/${{ github.repository }}.git
         run: build/registry/rules-registry push-to-oci registry.yaml ${{github.ref_name}}
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
```
LucaGuerra, line 46: Note for the reviewer: we could write this in plain text but I prefer a secret. WDYT?

LucaGuerra:
/hold

until we sort out those two points, then I'll add the secret or var if necessary and merge

.github/workflows/release.yaml
LucaGuerra: we can use a secret for now, it could be changed in the future

build/registry/s3.go
LucaGuerra: I don't think so, based on this: https://github.com/falcosecurity/test-infra/blob/be8f54f98393f743a0374b664f36f717c5505fb9/images/build-plugins/build-and-publish.sh#L41

leogr (approved):
:+1:

PR #20, opened 2023-01-25
new(rules): introduce rule to detect drop and execute pattern in cont…

loresuso:
…ainers

Signed-off-by: Lorenzo Susini <susinilorenzo1@gmail.com>

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/area rules

> /area registry

> /area build

> /area documentation

Moving the discussion here from https://github.com/falcosecurity/falco/pull/2306

Fixes #

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -17,7 +17,7 @@
 
 # Starting with version 8, the Falco engine supports exceptions.
 # However the Falco rules file does not use them by default.
-- required_engine_version: 13
+- required_engine_version: 16
 
 # Currently disabled as read/write are ignored syscalls. The nearly
 # similar open_write/open_read check for files being opened for
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -490,6 +490,20 @@
     WARNING
   tags: [host, container, filesystem, mitre_discovery, T1546.004]
 
+- rule: Drop and execute new binary in container
+  desc:
+    Detect if an executable not belonging to the base image of a container is being executed.
+    The drop and execute pattern can be observed very often after an attacker gained an initial foothold.
+    is_exe_upper_layer filter field only applies for container runtimes that use overlayfs as union mount filesystem.
+  condition: spawned_process and container and proc.is_exe_upper_layer=true
+  output: > 
+    Executing binary not part of base image (user=%user.name user_loginuid=%user.loginuid user_uid=%user.uid comm=%proc.cmdline exe=%proc.exe container_id=%container.id
+    image=%container.image.repository proc.name=%proc.name proc.sname=%proc.sname proc.pname=%proc.pname proc.aname[2]=%proc.aname[2] exe_flags=%evt.arg.flags
+    proc.exe_ino=%proc.exe_ino proc.exe_ino.ctime=%proc.exe_ino.ctime proc.exe_ino.mtime=%proc.exe_ino.mtime proc.exe_ino.ctime_duration_proc_start=%proc.exe_ino.ctime_duration_proc_start
+    proc.exepath=%proc.exepath proc.cwd=%proc.cwd proc.tty=%proc.tty container.start_ts=%container.start_ts proc.sid=%proc.sid proc.vpgid=%proc.vpgid evt.res=%evt.res)
+  priority: CRITICAL 
+  tags: [container, mitre_persistence, TA0003]
+
 - macro: user_known_cron_jobs
   condition: (never_true)
 
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -490,12 +490,22 @@
     WARNING
   tags: [host, container, filesystem, mitre_discovery, T1546.004]
 
+# List of allowed container images that are known to execute binaries not part of their base image.
+# Users can use this list to better tune the rule below (i.e reducing false positives) by considering their workloads, 
+# since this requires application specific knowledge.
+- list: known_drop_and_execute_containers
+  items: []
+
 - rule: Drop and execute new binary in container
   desc:
     Detect if an executable not belonging to the base image of a container is being executed.
     The drop and execute pattern can be observed very often after an attacker gained an initial foothold.
     is_exe_upper_layer filter field only applies for container runtimes that use overlayfs as union mount filesystem.
-  condition: spawned_process and container and proc.is_exe_upper_layer=true
+  condition: >
+    spawned_process
+    and container
+    and proc.is_exe_upper_layer=true 
+    and not container.image.repository in (known_drop_and_execute_containers)
   output: > 
     Executing binary not part of base image (user=%user.name user_loginuid=%user.loginuid user_uid=%user.uid comm=%proc.cmdline exe=%proc.exe container_id=%container.id
     image=%container.image.repository proc.name=%proc.name proc.sname=%proc.sname proc.pname=%proc.pname proc.aname[2]=%proc.aname[2] exe_flags=%evt.arg.flags


rules/falco_rules.yaml
```diff
@@ -3333,6 +3333,17 @@
...
+- rule: Drop and execute new binary in container
+  desc:
+    Detect if an executable not belonging to the base image of a container is being executed.
+    The drop and execute pattern can be observed very often after an attacker gained an initial foothold.
+    is_exe_upper_layer filter field only applies for container runtimes that use overlayfs as union mount filesystem.
+  condition: >
+    spawned_process and container and proc.is_exe_upper_layer=true
+  output: Executing binary not part of base image (user=%user.name user_loginuid=%user.loginuid comm=%proc.cmdline exe=%proc.exe container_id=%container.id image=%container.image.repository, proc.name=%proc.name, proc.sname=%proc.sname, proc.pname=%proc.pname, proc.aname[2]=%proc.aname[2], exe_flags=%evt.arg.flags, proc.exe_ino=%proc.exe_ino, proc.exe_ino.ctime=%proc.exe_ino.ctime, proc.exe_ino.mtime=%proc.exe_ino.mtime proc.exe_ino.ctime_duration_proc_start=%proc.exe_ino.ctime_duration_proc_start, container.start_ts=%container.start_ts)
+  priority: CRITICAL 
+  tags: [container, mitre_execution]
```
incertum, line 3345: Lorenzo, would Mitre TTP [T1204.002](https://attack.mitre.org/techniques/T1204/002/) seem fitting? We are in the transitioning period where it's now added as tag so that end users have even more context around the threat a particular detection attempts to address, see new [rules overview doc](https://github.com/falcosecurity/rules/blob/main/rules_inventory/rules_overview.md).

rules/falco_rules.yaml
```diff
@@ -3333,6 +3333,17 @@
...
   tags: [host, container, mitre_execution, mitre_defense_evasion, T1036.005, T1059.004]
 
+- rule: Drop and execute new binary in container
+  desc:
+    Detect if an executable not belonging to the base image of a container is being executed.
+    The drop and execute pattern can be observed very often after an attacker gained an initial foothold.
+    is_exe_upper_layer filter field only applies for container runtimes that use overlayfs as union mount filesystem.
+  condition: >
+    spawned_process and container and proc.is_exe_upper_layer=true
+  output: Executing binary not part of base image (user=%user.name user_loginuid=%user.loginuid comm=%proc.cmdline exe=%proc.exe container_id=%container.id image=%container.image.repository, proc.name=%proc.name, proc.sname=%proc.sname, proc.pname=%proc.pname, proc.aname[2]=%proc.aname[2], exe_flags=%evt.arg.flags, proc.exe_ino=%proc.exe_ino, proc.exe_ino.ctime=%proc.exe_ino.ctime, proc.exe_ino.mtime=%proc.exe_ino.mtime proc.exe_ino.ctime_duration_proc_start=%proc.exe_ino.ctime_duration_proc_start, container.start_ts=%container.start_ts)
```
incertum, line 3343: - maybe also`%user.uid`
- highly recommending also adding `%proc.exepath` and `%proc.tty` so see if it was more from an interactive session and lastly `%evt.res`
- also putting `%proc.sid` and `%proc.vpgid` up for discussion also useful for sessionization of several rules

rules/falco_rules.yaml
```diff
@@ -3333,6 +3333,17 @@
   priority: WARNING
   tags: [host, container, mitre_execution, mitre_defense_evasion, T1036.005, T1059.004]
 
+- rule: Drop and execute new binary in container
+  desc:
+    Detect if an executable not belonging to the base image of a container is being executed.
+    The drop and execute pattern can be observed very often after an attacker gained an initial foothold.
+    is_exe_upper_layer filter field only applies for container runtimes that use overlayfs as union mount filesystem.
+  condition: >
+    spawned_process and container and proc.is_exe_upper_layer=true
```
incertum, line 3342: In terms of filtering perhaps check if it would make sense to also filter by `%proc.exe_ino.ctime_duration_pidns_start` (it is the time delta in nanoseconds between container start and ctime of the exe_file)? For example, exclude the first n nanoseconds of the lifetime of a container as containers may pull artifacts or do bunch of symlinking right after start up (symlinks also create a new inode aka a new executable in our definition / rule here).

incertum (commented):
@loresuso it goes without saying what a game changer this rule is. Knocking out an entire class of vulnerabilities in one simple rule is 🔥 !!! And our journey won't end here 😉 .

Also thanks for adding limitations into the desc, as also bind mounts wouldn't be on the upper layer.

rules/falco_rules.yaml
loresuso: Hello @incertum! Yes, I have noticed that we are putting a tag with a fitting MITRE technique, and this indeed was something that I was going to ask you. Not an expert on that framework yet, but I believe T1059 is not really our case here. I will look into this problem and see if we can find something more fitting!

rules/falco_rules.yaml
loresuso: Completely agree with everything on this. 
On the very last point, I didn't see any shadowing for the rule, but however given its importance is better to be safe and put it somewhere at the top of the file. I am doing this!

rules/falco_rules.yaml
loresuso: This is something I considered. I would say let's start with not cutting the beginning of container life. I still believe in most use cases containers will be immutable. And in any case, how to choose the threshold `n`? What we can do instead, is allow known containers to pull and execute artifacts within n nanoseconds by means of a whitelist that we can introduce later to fine-tune the rule if it becomes too noisy :)

Moreover on the symlink point: `exe_upper_layer` as you know retrieves the inode of the executable from the `mm_struct->exe_file` from the process descriptor (task_struct). This is something that the kernel already resolved for us, so what we are taking here in case of symlink is the real inode the symlink is pointing to. So a symlink to something that is part of the base image will not cause the rule to trigger! I need to double check this, but I am pretty sure it works like this

rules/falco_rules.yaml
incertum: Thanks for checking which TTP may be most fitting!

rules/falco_rules.yaml
incertum: re tuning agreed, also think for the most part containers should be immutable and when they are not the tuning may be too organization and application specific hence similar to other rules, the end user can customize the rule. And what you suggested with "known containers to pull and execute artifacts within n nanoseconds" sounds great. Maybe instead of adding tuning already just list a few options as comment above? Consider this feedback [nit], no need to address. Finding the right thresholds and such yeah that is the tricky part. Also I forgot to implement something like `ctime - mtime` ... a parameter I noticed can contain high value info as well.

re symlinks at runtime let me double check as well.

incertum: > Moreover on the symlink point: `exe_upper_layer` as you know retrieves the inode of the executable from the `mm_struct->exe_file` from the process descriptor (task_struct). This is something that the kernel already resolved for us, so what we are taking here in case of symlink is the real inode the symlink is pointing to. So a symlink to something that is part of the base image will not cause the rule to trigger! I need to double check this, but I am pretty sure it works like this


correct, also just verified that if you dummy symlink 2 binaries that were already in the container -> not in upper layer and what you describe is how you can detect for example ` /bin/uname -> /bin/new_malicious`, but your exepath would be `/bin/uname`, see https://github.com/falcosecurity/libs/pull/705#issuecomment-1312627375 I think that's what would be intended to be detected in such cases right?

rules/falco_rules.yaml
loresuso: Mm, I am checking the matrix both for linux and containers (https://attack.mitre.org/matrices/enterprise/linux/) and I am not seeing something really fitting honestly. I am also wondering: this could be part of the execution and persistence category in my opinion. Execution is pretty obvious, but if I think about that, the attacker here would already have gained initial access, found a way to drop the new binary (maybe after gaining RCE, which would already be execution) and only later execute a binary of its choice. However, also under persistence I cannot find something that describe this well

rules/falco_rules.yaml
loresuso: Strongly agreed with `ctime - mtime`, this could be particularly relevant!
Not sure what you mean by the second point, but what I was saying is that no matter how you deal with symlinks, if you symlink to something that was in the base image the rule doesn't have to trigger (and it does not), else it has to trigger (this case is correctly handled). Of course, using symlinks the exepath could potentially be wrong due to the fact that we retrieve it from userspace and use it unresolved (at least for now!) 🙂 Did I get your point right or not?

rules/falco_rules.yaml
incertum: Maybe just link to the generic TA since I mean it kind of is an awesome generic detection https://attack.mitre.org/tactics/TA0011/, https://attack.mitre.org/tactics/TA0002/ -> done this at other places that way too.

rules/falco_rules.yaml
incertum: @loresuso re your summary yup that's about it what I meant. It may seem obvious to us, but for some folks reading this it may improve the clarity of what this detection captures.

Re `ctime - mtime` more longer term I just opened this https://github.com/falcosecurity/falco/issues/2403 so maybe in the future we will be more flexibility for such cases without needed to introduce a new filtercheck field in the source.


---

Back to the detection: I think the tuning really will be very application and organization specific - just like many of the other rules. The biggest tuning win likely is profiled containers/apps combined w/ process lineage information as we discussed before, so maybe this as template lists or hint could make sense, and then it can be further tuned after field experience. It definitely won't explode the data pipeline for sure would say.

rules/falco_rules.yaml
```diff
@@ -490,6 +490,20 @@
     WARNING
   tags: [host, container, filesystem, mitre_discovery, T1546.004]
 
+- rule: Drop and execute new binary in container
+  desc:
+    Detect if an executable not belonging to the base image of a container is being executed.
+    The drop and execute pattern can be observed very often after an attacker gained an initial foothold.
+    is_exe_upper_layer filter field only applies for container runtimes that use overlayfs as union mount filesystem.
+  condition: spawned_process and container and proc.is_exe_upper_layer=true
```
Kaizhe, line 498: This is awesome!  @loresuso you should write a blog about this rule.

rules/falco_rules.yaml
```diff
@@ -490,6 +490,20 @@
...
+  output: >
```
Kaizhe, line 499: I think the output is a bit too much.

rules/falco_rules.yaml
```diff
@@ -490,6 +490,20 @@
...
+    The drop and execute pattern can be observed very often after an attacker gained an initial foothold.
+    is_exe_upper_layer filter field only applies for container runtimes that use overlayfs as union mount filesystem.
+  condition: spawned_process and container and proc.is_exe_upper_layer=true
+  output: > 
+    Executing binary not part of base image (user=%user.name user_loginuid=%user.loginuid user_uid=%user.uid comm=%proc.cmdline exe=%proc.exe container_id=%container.id
+    image=%container.image.repository proc.name=%proc.name proc.sname=%proc.sname proc.pname=%proc.pname proc.aname[2]=%proc.aname[2] exe_flags=%evt.arg.flags
+    proc.exe_ino=%proc.exe_ino proc.exe_ino.ctime=%proc.exe_ino.ctime proc.exe_ino.mtime=%proc.exe_ino.mtime proc.exe_ino.ctime_duration_proc_start=%proc.exe_ino.ctime_duration_proc_start
+    proc.exepath=%proc.exepath proc.cwd=%proc.cwd proc.tty=%proc.tty container.start_ts=%container.start_ts proc.sid=%proc.sid proc.vpgid=%proc.vpgid evt.res=%evt.res)
+  priority: CRITICAL 
+  tags: [container, mitre_execution]
```
Kaizhe, line 505: `mitre_persistence`

rules/falco_rules.yaml
loresuso: Thank you @Kaizhe, really appreciated! 🙏 
Yes, I was thinking about it, I will definitely do it!

rules/falco_rules.yaml
loresuso: What would you drop from here?

rules/falco_rules.yaml
loresuso: I've added mitre persistence and its corresponding TA as per Kaizhe comment, what do you think? :)

rules/falco_rules.yaml
loresuso: Nice! 
Added a list and commented on it so that users can already use it if they need. Thanks!

incertum:
@loresuso almost forgot: We need to bump the `required_engine_version` to I think 15 is the current version, see also https://github.com/falcosecurity/rules/issues/25 to have this as reminder in template.

FedeDP (approved):
/approve

FedeDP:
This is an incredibly feat @loresuso ! Nice work!

PR #24, opened 2023-02-07
Ignore regular writes by cupsd to its own files

petterreinholdtsen:
The CUPS daemon update its /etc/cups/subscriptions.conf file regularly, no need to create noise in the logs about this.

/kind bug
/kind cleanup
/kind feature
/area rules

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -1172,6 +1172,17 @@
 - macro: calico_node
   condition: (container.image.repository endswith calico/node and proc.name=calico-node)
 
+# Return always_true from this macro to get warnings about the cups daemons
+# regular updates in /etc/cups/.
+- macro: cupsd_state_updates
+  condition: never_true
+
+- macro: cupsd_state_writing
+  condition: >
+    proc.name = cupsd and
+    fd.name startswith /etc/cups/ and
+    not cupsd_state_updates
+
 - macro: write_etc_common
   condition: >
     etc_dir and evt.dir = < and open_write
@@ -1278,6 +1289,7 @@
     and not avinetworks_supervisor_writing_ssh
     and not multipath_writing_conf
     and not calico_node
+    and not cupsd_state_writing
 
 - rule: Write below etc
   desc: an attempt to write to any file below /etc


rules/falco_rules.yaml
```diff
@@ -1172,6 +1172,23 @@
...
+- macro: cupsd_state_writing
+  condition: >
+    proc.name = cupsd and
+    fd.name in (
+       /etc/cups/subscriptions.conf,
+       /etc/cups/subscriptions.conf.N,
+       /etc/cups/subscriptions.conf.O,
+       /etc/cups/printers.conf,
+       /etc/cups/printers.conf.N,
+       /etc/cups/printers.conf.O
```
loresuso, line 1189: Hello @petterreinholdtsen, I am not familiar with cupsd but just wanted to ask a question, are we sure that these are the only file it is writing? After installation, I am seeing in `/etc/cups` that there are also other files and directories. 
Maybe we can con consider using `startswith /etc/cups` instead of `in`. Just a question, in any case sgtm

petterreinholdtsen:
[Lorenzo Susini]
> Hello @petterreinholdtsen, I am not familiar with cupsd but just
> wanted to ask a question, are we sure that these are the only file it
> is writing? After installation, I am seeing in `/etc/cups` that there
> are also other files and directories.

The files listed are the ones that I am aware of are automatically
updated without human interaction.  Other files (as well as these) are
updated by cupsd when printers and settings are configured, but I
suspect a log entry might be in order when that happen?

> Maybe we can con consider using `startswith /etc/cups` instead of
> `in`. Just a question, in any case sgtm

It depend on how fine grained the rule should be.  Everything under
/etc/cups/ "belong" to cupsd, so it would not be surprising that it
updated one of its own files.

-- 
Happy hacking
Petter Reinholdtsen

loresuso:
Since this rule has been noisy historically, I'd use `startswith`. Moreover, I see no security implications and as you said "not surprising" cupsd writing to its own files. This is my opinion, let's wait for other people to join the conversation

petterreinholdtsen:
[Lorenzo Susini]
> Since this rule has been noisy historically, I'd use
> `startswith`. Moreover, I see no security implications and as you said
> "not surprising" cupsd writing to its own files. This is my opinion,
> let's wait for other people to join the conversation

I have no problem with a less specific test.  Patch updated to use
fd.name startswith /etc/cups/ instead.

-- 
Happy hacking
Petter Reinholdtsen

rules/falco_rules.yaml
```diff
@@ -1172,6 +1172,17 @@
 - macro: calico_node
   condition: (container.image.repository endswith calico/node and proc.name=calico-node)
 
+# Return always_true from this macro to get warnings about the cups daemons
+# regular updates in /etc/cups/.
+- macro: cupsd_state_updates
```
darryk10, line 1177: I'm not sure if this macro is actually useful. In the ruleset we have already never_true macros and I would avoid adding a new one if not strictly required.

petterreinholdtsen:
[schie]
> I'm not sure if this macro is actually useful. In the ruleset we have
> already never_true macros and I would avoid adding a new one if not
> strictly required.

The macro is there to allow a site to control the behaviour using local
overrides.

-- 
Vennlig hilsen
Petter Reinholdtsen

petterreinholdtsen:
It is unclear to me what more I can do to make this acceptable into the
default rule set.  Should I drop the option to allow the override and
always ignore changes in /etc/cups/?

I have rebased the change on top of the current main branch.

-- 
Happy hacking
Petter Reinholdtsen

PR #26, opened 2023-02-13
Added dnf-automatic to list of rpm_binaries

petterreinholdtsen:
The process is called by systemd regularly and write to /var/lib/rpm/rpmdb.sqlite-shm.  This will quiet down the event log on RHEL 9 and possibly other RHEL versions.

/kind bug
/area rules

Signed-off-by: Petter Reinholdtsen <pere@hungry.com>

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -218,7 +218,7 @@
 # The explicit quotes are needed to avoid the - characters being
 # interpreted by the filter expression.
 - list: rpm_binaries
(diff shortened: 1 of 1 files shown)


loresuso (approved):

petterreinholdtsen:
Is there a timeline on when this change can make it into the published branch?

leogr:
> Is there a timeline on when this change can make it into the published branch?

I will merge this soon, but we have no plan for the release date yet. In the worst case, it will be part of the next Falco release (in May).

leogr (approved):

PR #34, opened 2023-02-24
update(falco_rules): disable three rules by default

jasondellaluce:
**What type of PR is this?**

/kind cleanup

/area rules

This sets three rules as disabled by default, each for a different reason:
- `Contact EC2 Instance Metadata Service From Container`: without custom overrides, the rule contains the unrolled condition `container and not container`, which means that the rule never triggers at default and just wastes evaluation time.
- `Outbound Connection to C2 Servers`: without custom overrides, the condition will require one of `fd.sip` and `fd.sip.name` to be contained in empty lists, which is impossible, meaning that the rule never triggers at default and just wastes evaluation time.
- `Java Process Class File Download`: This matches the `recvfrom` syscall, which is IO-related and performance-heavy. The recent Falco developments will automatically exclude those syscalls if the `-A` option is not used, which means that this rule will never trigger by default in the next Falco release.

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3120,6 +3120,7 @@
      (fd.sip.name in (c2_server_fqdn_list)))
   output: Outbound connection to C2 server (c2_domain=%fd.sip.name c2_addr=%fd.sip command=%proc.cmdline connection=%fd.name user=%user.name user_loginuid=%user.loginuid container_id=%container.id image=%container.image.repository)
   priority: WARNING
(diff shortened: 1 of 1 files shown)


loresuso (approved):
Sounds good!
/approve

darryk10 (approved):
totally agree with this. users can enable those rules is specifically interested in the use case adding values to lists and changing macros.

FedeDP (approved):
/approve

PR #37, opened 2023-03-15
fix(rules): fix ptrace attach and injection rule

loresuso:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/area rules

> /area registry

> /area build

> /area documentation

This rule was built on top of a bug, see https://github.com/falcosecurity/libs/pull/970
Once that one will be merged and Falco can use it, we should merge this PR as well.
For the same reason, some other rules stopped working, like https://github.com/falcosecurity/rules/blob/main/rules/falco_rules.yaml#L2950 because the flag was not retrieved as string anymore. This makes me think we need a better CI :) 

Fixes #

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2949,7 +2949,7 @@
 
 - rule: Packet socket created in container
   desc: Detect new packet socket at the device driver (OSI Layer 2) level in a container. Packet socket could be used for ARP Spoofing and privilege escalation(CVE-2020-14386) by attacker.
-  condition: evt.type=socket and evt.arg[0]=AF_PACKET and container and not proc.name in (user_known_packet_socket_binaries)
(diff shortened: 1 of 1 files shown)


darryk10:
hi @loresuso, thanks for the that and the rule seems more accurate then before so that's great.
Reviewing the PR isn't clear the other rules that stopped working since the link point to a macro https://github.com/falcosecurity/rules/blob/main/rules/falco_rules.yaml#L2950. Can you clarify that pls?
Thanks

darryk10:
noted! thanks.
this LGTM.

darryk10 (dismissed):

loresuso:
I have added other commits to:
- make the "Packet socket created in container" work again. This commit can be cherry picked if we want to release a patched ruleset
- revert the commit before
- let the rule use `contains` operator is better because of the way things are extracted. This is what we will keep after the fix got it.

PR #38, opened 2023-03-16
Update falco_rules.yaml

yogirajk:
Priority Modifications

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /area rules

> /area registry

> /area build

> /area documentation

Fixes #

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -483,7 +483,7 @@
      fd.name in (shell_config_files) or
      fd.directory in (shell_config_directories)) and
     (not proc.name in (shell_binaries))
-  enabled: false
+  enabled: true
   output: >
     a shell configuration file was read by a non-shell program (user=%user.name user_loginuid=%user.loginuid command=%proc.cmdline pid=%proc.pid file=%fd.name container_id=%container.id image=%container.image.repository)
   priority:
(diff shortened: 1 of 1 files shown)


leogr (commented):
Could you elaborate more on why you are proposing these changes?

leogr:
cc @darryk10 @loresuso

rules/falco_rules.yaml
```diff
@@ -2497,11 +2497,11 @@
   desc: Detect attempts to use K8s NodePorts from a container
   condition: (inbound_outbound) and fd.sport >= 30000 and fd.sport <= 32767 and container and not nodeport_containers
   output: Unexpected K8s NodePort Connection (command=%proc.cmdline pid=%proc.pid connection=%fd.name container_id=%container.id image=%container.image.repository)
-  priority: NOTICE
+  priority: CRITICAL
   tags: [network, k8s, container, mitre_persistence, T1205.001]
 
 - list: network_tool_binaries
-  items: [nc, ncat, nmap, dig, tcpdump, tshark, ngrep, telnet, mitmproxy, socat, zmap]
+  items: [nc, ncat, nmap, tcpdump, tshark, ngrep, telnet, mitmproxy, socat, zmap]
```
darryk10, line 2504: this is definitely a good point and we should reduce the false positive as much as possible.
Wondering if there might be a specific whitelist we can apply for dig used in OpenShift since this is the major use case.

darryk10 (changes requested):
Hi @yogirajk thanks for your contribution.
I see this PR contains different changes which some of those might be be related to what you mentioned in the PR. 
Could you mind open different PRs for the changes provide. The dig change definitely requires a separate PR to better discuss rather then putting it together with priority changes.
There are also changes in enabled flags which I'm not sure is related to the changes proposed. If you want to submit changes for this flag pls open a separate PR.

rules/falco_rules.yaml
yogirajk: sure.. I will create separate PR for this... Closing this pull request .. For now only this change will work for me

PR #39, opened 2023-03-20
update falco_rules for 

yogirajk:
Fixes #

We have noticed that one of the rules, which includes "dig" in the list of suspicious network tools, may produce false positive results in OpenShift. This is because the DNS pods in OpenShift frequently use "dig, for DNS resolution" and therefore it should not be considered a suspicious network tool in this particular environment.

Adding to this, we have also set a Kubernetes response engine on this rule for the tools like nmap, nc which is also restarting our openshift DNS pods.

Here is the original code block for reference:
----------------------------------
list: network_tool_binaries
items: [nc, ncat, dig, nmap, tcpdump, tshark, ngrep, telnet, mitmproxy, socat, zmap]
----------------------------------

### DIFF
--- a/rules/custom_rules.yaml
+++ b/rules/custom_rules.yaml
@@ -1,0 +1,2160 @@
+customRules:
+  custom_k8s_rules.yaml: |-
+    - rule: Launch Suspicious Network Tool in Container
+      desc: Detect network tools launched inside container
+      condition: >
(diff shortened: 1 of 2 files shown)


darryk10:
Hi @yogirajk do you mind remove the custom rules files since not related to this PR?
In addition you need to add the Signed-off-by in the commit and PR so that it can be accepted.

PR #43, opened 2023-03-31
cleanup(rules)!: refactor kernel module rule

incertum:
**What type of PR is this?**

/kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

/area rules

> /area registry

> /area build

> /area documentation

Rule "Linux Kernel Module Injection Detected" is limited in scope.
A major refactor of Falco now exposes each syscall Falco's libs supports (including `init_module`) to the end user :)
Official support starts with Falco 0.35.0.

https://github.com/falcosecurity/falco/issues/2443#issuecomment-1471289452

Fixes #

CC @loresuso @LucaGuerra @jasondellaluce 

```
cleanup(rules)!: refactor kernel module rule
```

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -17,7 +17,7 @@
 
 # Starting with version 8, the Falco engine supports exceptions.
 # However the Falco rules file does not use them by default.
-- required_engine_version: 16
+- required_engine_version: 17
 
 # Currently disabled as read/write are ignored syscalls. The nearly
 # similar open_write/open_read check for files being opened for
@@ -85,6 +85,9 @@
 
 - macro: chmod
   condition: (evt.type in (chmod, fchmod, fchmodat) and evt.dir=<)
+
+- macro: init_module
+  condition: (evt.type=init_module and evt.dir=<)
 
 # File categories
 - macro: bin_dir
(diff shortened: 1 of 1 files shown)


rules/falco_rules.yaml
```diff
@@ -3150,10 +3153,15 @@
...
 
+# init_module syscall is available since Falco 0.35.0
+# rule coverage now extends to modprobe usage
 - rule: Linux Kernel Module Injection Detected
   desc: Detect kernel module was injected (from container).
-  condition: spawned_process and container and proc.name=insmod and not proc.args in (white_listed_modules) and thread.cap_effective icontains sys_module
-  output: Linux Kernel Module injection using insmod detected (user=%user.name user_loginuid=%user.loginuid parent_process=%proc.pname module=%proc.args %container.info image=%container.image.repository:%container.image.tag)
+  condition: ((spawned_process and proc.name=insmod) or init_module)
+    and container
+    and not proc.args in (white_listed_modules)
```
loresuso, line 3162: Nice, thanks for noticing!
Btw, I am thinking that maybe it's time to change this rule and make it even better.
I would like to drop the part of the condition relying on `insmod`, and use instead the related syscalls, which is of course `init_module` but also `finit_module` (which is what my `insmod` is using). wdyt? :)

incertum: Thank you @loresuso this is great! Done :)

loresuso: Nice! One last thing is either throw away the list `white_listed_modules` cause now it should be unused.
I'd add an `allowed_container_images_loading_kernel_module` list and add it to the condition with `and not` and it's perfect for me!

rules/falco_rules.yaml
```diff
@@ -3147,13 +3150,14 @@
   enabled: false
   tags: [host, container, network, mitre_command_and_control, TA0011]
 
-- list: white_listed_modules
```
Kaizhe, line 3150: who do we remove the whitelist ?

loresuso: For me now given that the rule is changing, it makes sense having a whitelist on *container images* that can load modules

Kaizhe: did you add it in the condition

loresuso: no, that's what I was asking (see above), let's wait for Melissa :)

Kaizhe: with the whitelisted images (e.g. falco), I think we're good to go.

incertum: Added, this is great feedback, ty :)

loresuso (approved):
Nice! /approve

jasondellaluce (approved):
/approve

jasondellaluce:
Note for us: renaming the macro can represent a breaking change. Let's keep this in mind for the next falco_rules release.

Kaizhe:
> Note for us: renaming the macro can represent a breaking change. Let's keep this in mind for the next falco_rules release.

Maybe let's revert the macro, we should probably give a heads up in the release notes (saying this macro will be deprecated in release 3.8 or 9)

PR #45, opened 2023-04-01
new(rules): add umount macro

incertum:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/area rules

> /area registry

> /area build

> /area documentation

add new `umount` macro

Fixes #

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -74,6 +74,11 @@
 - macro: modify
   condition: (rename or remove)
 
+# %evt.arg.flags available for evt.dir=>, but only for umount2
+# %evt.arg.name is path and available for evt.dir=<
+- macro: umount
+  condition: (evt.type in (umount, umount2))
+
 - macro: spawned_process
   condition: (evt.type in (execve, execveat) and evt.dir=<)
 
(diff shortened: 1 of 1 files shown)


rules/falco_rules.yaml
```diff
@@ -74,6 +74,11 @@
 - macro: modify
   condition: (rename or remove)
 
+# %evt.arg.flags available for evt.dir=>, but only for umount2
+# %evt.arg.name is path and available for evt.dir=<
+- macro: umount
```
loresuso, line 79: is this for an upcoming rule or for addressing https://github.com/falcosecurity/falco/issues/2443#issuecomment-1471289452? in that case, do we want macros also for the event types (e.g pipe,pipe2) already put in place?

jasondellaluce:
My only concern of adding macros that are not used in rules is that Falco will return the "XXX macro is not used" warning when processing the ruleset for validation.

Andreagit97:
Agree with Jason here, we could comment it and leave it as a hint for users, WDYT?

jasondellaluce (approved):
/approve

PR #48, opened 2023-04-10
adds rule for detetcing system recovery features

pandyamarut:
Signed-off-by: pandyamarut [pandyamarut@gmail.com](mailto:pandyamarut@gmail.com) ([cloudDefense.ai](https://www.clouddefense.ai/))
**What type of PR is this?**

/kind feature

 /area rules

adds a rule to detect disabling system recovery feature.

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3398,6 +3398,29 @@
   priority: WARNING
   tags: [host, container, mitre_execution, mitre_defense_evasion, T1036.005, T1059.004]
 
+#The given rule is designed to detect suspicious activities related to disabling system recovery features. 
+#The rule looks for specific commands being executed such as rm -rf, chattr +i, mkfs.ext4 -E nodiscard, mount remount,ro, 
+#and disabling certain services such as systemd-backlight@.service, apport.service, rescue.service, emergency.service, and recovery-mode.service. 
+#https://attack.mitre.org/techniques/T1490/
+#Mitre Impact: Inhibit system recovery
+
+- rule: Disable Recovery Features
+  desc: Detects disabling system recovery features by deleting or disabling services and commands.
+  condition: >
+      (spawned_process and proc.name = "rm" and proc.args contains "-rf") or
+      (spawned_process and proc.name = "chattr" and proc.args contains "+i") or
+      (spawned_process and proc.name = "mkfs.ext4" and proc.args contains "-E nodiscard") or
+      (spawned_process and proc.name = "mount" and proc.args contains "remount,ro") or
+      (spawned_process and proc.name = "systemctl" and proc.args contains "disable systemd-backlight@.service") or 
+      (spawned_process and proc.name = "systemctl" and proc.args contains "disable apport.service") or
+      (spawned_process and proc.name = "systemctl" and proc.args contains "disable rescue.service") or
+      (spawned_process and proc.name = "systemctl" and proc.args contains "disable emergency.service") or
+      (spawned_process and proc.name = "systemctl" and proc.args contains "disable recovery-mode.service")
+  enabled: true
+  output: "Disabling recovery features so that system becomes non recoverable in case of failure."
+  priority: CRITICAL
+  tags: [host, container, recovery, mitre_impact, inhibit_system_recovery, T1490]
+
 # Application rules have moved to application_rules.yaml. Please look
 # there if you want to enable them by adding to
 # falco_rules.local.yaml.
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3404,20 +3404,26 @@
 #https://attack.mitre.org/techniques/T1490/
 #Mitre Impact: Inhibit system recovery
 
+- list: recovery_files
+  items: [/etc/default/grub, /etc/inittab, /usr/lib/systemd/system/rescue.service]
+
 - rule: Disable Recovery Features
   desc: Detects disabling system recovery features by deleting or disabling services and commands.
   condition: >
-      (spawned_process and proc.name = "rm" and proc.args contains "-rf") or
-      (spawned_process and proc.name = "chattr" and proc.args contains "+i") or
-      (spawned_process and proc.name = "mkfs.ext4" and proc.args contains "-E nodiscard") or
-      (spawned_process and proc.name = "mount" and proc.args contains "remount,ro") or
-      (spawned_process and proc.name = "systemctl" and proc.args contains "disable systemd-backlight@.service") or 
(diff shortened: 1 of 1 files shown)


rules/falco_rules.yaml
```diff
@@ -3398,6 +3398,29 @@
...
+#The given rule is designed to detect suspicious activities related to disabling system recovery features. 
+#The rule looks for specific commands being executed such as rm -rf, chattr +i, mkfs.ext4 -E nodiscard, mount remount,ro, 
+#and disabling certain services such as systemd-backlight@.service, apport.service, rescue.service, emergency.service, and recovery-mode.service. 
+#https://attack.mitre.org/techniques/T1490/
+#Mitre Impact: Inhibit system recovery
+
+- rule: Disable Recovery Features
+  desc: Detects disabling system recovery features by deleting or disabling services and commands.
+  condition: >
+      (spawned_process and proc.name = "rm" and proc.args contains "-rf") or
```
incertum, line 3410: Some initial feedback: 

- We can use `spawned_process` first and then do the conditioning afterwards in a `spawned_process and ( () or () or () )`. This will reduce the AST expression and since the rules push down is by event type the overall rules evaluation will become more clear.
- Is this detections intended to run on the host only?
- Could some conditions around the file system be based on respective syscalls for increased robustness vs spawned process and command args?
- In the `output` we need the output fields, see existing rules for guidance, we can discuss which fields are most relevant.

pandyamarut: @incertum  Sure.

rules/falco_rules.yaml
```diff
@@ -3398,6 +3398,35 @@
...
+
+- list: recovery_files
+  items: [/etc/default/grub, /etc/inittab, /usr/lib/systemd/system/rescue.service]
+
+- rule: Disable Recovery Features
+  desc: Detects disabling system recovery features by deleting or disabling services and commands.
+  condition: >
+      (spawned_process) and 
+      ((open_write and fd.name in (recovery_files)) or
+      (proc.name = "rm" and proc.args contains "-rf") or
```
loresuso, line 3415: I'm always a little bit concerned when I see rules matching on proc.args. For instance, `rm -rf` operation can be accomplished also with `rm -r -f` or `rm -fr`. Still didn't find a way to solve this, apart from exhaustively write down all cases (doable with two flags, but imagine with 3 or 4..)

loresuso (commented):
I have to take a better look ar the systemctl services because, I don't know about them, but I'm leaving an early feedback anyway

rules/falco_rules.yaml
pandyamarut: @loresuso  Thanks for the comments. will make changes accordingly.

nidhisysdig: @pandyamarut - Keep us posted, when the changes are ready.

pandyamarut: @nidhisysdig pushed one commit.

darryk10:
Hi @pandyamarut I really like the use case. I'm a bit worried about the noise so I'm going to test as well to see if we can to come up with some whitelists if needed.

pandyamarut:
> Hi @pandyamarut I really like the use case. I'm a bit worried about the noise so I'm going to test as well to see if we can to come up with some whitelists if needed.

Hi @darryk10  Thanks. Looking forward to it.

loresuso:
Hello @pandyamarut, it looks ok to me now, but one last question, can you give me examples or use case for the services mentioned in the condition? What the attacker can achieve by disabling them? Just asking so to have a better picture of the rule.

rules/falco_rules.yaml
loresuso: Ok for the remove part. Now the concern is that we are using different event types, so we have to not create impossible conditions. What I would like to check is:
- the part that checks `spawned_process` must be paired with the `proc.name` parts of the condition
- the parts with open_write or remove must be ORed together

rules/falco_rules.yaml
```diff
@@ -3398,6 +3398,38 @@
...
+      ((open_write and fd.name in (recovery_files)) or
+      (remove and (fd.name in (critical_dir))) or
+      (proc.name = "chattr" and proc.args contains "+i") or
+      (proc.name = "mkfs.ext4" and proc.args contains "-E nodiscard") or
+      (proc.name = "mount" and proc.args contains "remount,ro") or
+      (proc.name = "systemctl" and proc.args contains "disable systemd-backlight@.service") or 
+      (proc.name = "systemctl" and proc.args contains "disable apport.service") or
+      (proc.name = "systemctl" and proc.args contains "disable rescue.service") or
+      (proc.name = "systemctl" and proc.args contains "disable emergency.service") or
+      (proc.name = "systemctl" and proc.args contains "disable recovery-mode.service"))
```
loresuso, line 3426: That would become:

### Condition rule 1
```
spawned_process and 
((proc.name = chattr and proc.args contains +i) or
(proc.name = "mkfs.ext4" and proc.args contains "-E nodiscard") or
<redacted, everything with proc.name here>)
```

### Condition rule 2
```
open_write and fd.name in (recovery_files)
```

### Condition rule 3 
```
remove and (fd.name in (critical_dir))
```
here we could also think to delete the macro and use operator `pmatch`, to match on the directory prefixes, and trigger if also files are deleted and not the directory as a whole

pandyamarut: @loresuso yep. we can do that. let me send these changes as well.

loresuso: so can we split the rule in three different ones? I prefer to have them separate once a rule takes into account many scenarios like this one.

loresuso:
Moreover, commits must follow the conventional commit format https://www.conventionalcommits.org/en/v1.0.0/
Can you please modify the commit messages? thanks in advance!

rules/falco_rules.yaml
```diff
@@ -3398,6 +3398,37 @@
...
+  desc: Detects disabling system recovery features by deleting or disabling services and commands.
+  condition: >
+      (spawned_process) and 
+      ((proc.name = "chattr" and proc.args contains "+i") or
+      (proc.name = "mkfs.ext4" and proc.args contains "-E nodiscard") or
+      (proc.name = "mount" and proc.args contains "remount,ro") or
+      (proc.name = "systemctl" and proc.args in (recovery_proc_args)) or
+      (open_write and fd.name in (recovery_files)) or
+      (remove and (fd.name pmatch (critical_dir))))
+  enabled: true
```
darryk10, line 3426: Can you also remove the enabled:true?
If it's true we don't need to specify this.

darryk10 (changes requested):

PR #50, opened 2023-04-25
adds rule for detetcing getting information about open app'n windows.

ashutoshmore658:
Signed-off-by:ashutoshmore658 ashutoshmore658@gmail.com (cloudDefense.ai)
**What type of  PR is this?**

> /kind feature

> /area rules

Adds a rule to detect special kind of malicious system activity which tries to gather system information by getting information about open application windows in a system.

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3398,14 +3398,14 @@
   priority: WARNING
   tags: [host, container, mitre_execution, mitre_defense_evasion, T1036.005, T1059.004]
  
-- list: open_appn_windows
+- list: open_appln_windows
   items: ["xwininfo", "wmctrl", "xprop", "xlsclients", "xdotool", "xrestop", "xwininfo", "xdpyinfo", "glxinfo", "xrandr", "xset", "xdriinfo", "xhost", "xev", "xdotool", "xsetroot"]
    
 - rule: Get Information About Open Application Windows
   desc: Detects attempts to get information about open application windows
   condition: >
     spawned_process and 
-    (proc.name in (open_appn_windows))
+    (proc.name in (open_appln_windows))
   enabled: true
   output: "Get info about open windows in system detected (proc.cmdline=%proc.cmdline connection=%fd.name user.name=%user.name user.loginuid=%user.loginuid container.id=%container.id evt.type=%evt.type     evt.res=%evt.res proc.pid=%proc.pid proc.cwd=%proc.cwd proc.ppid=%proc.ppid proc.pcmdline=%proc.pcmdline proc.sid=%proc.sid proc.exepath=%proc.exepath user.uid=%user.uid user.loginname=%user.loginname group.gid=%group.gid group.name=%group.name container.name=%container.name image=%container.image.repository)"
   priority: WARNING
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3397,11 +3397,15 @@
   output: "File execution detected from /dev/shm (proc.cmdline=%proc.cmdline connection=%fd.name user.name=%user.name user.loginuid=%user.loginuid container.id=%container.id evt.type=%evt.type evt.res=%evt.res proc.pid=%proc.pid proc.cwd=%proc.cwd proc.ppid=%proc.ppid proc.pcmdline=%proc.pcmdline proc.sid=%proc.sid proc.exepath=%proc.exepath user.uid=%user.uid user.loginname=%user.loginname group.gid=%group.gid group.name=%group.name container.name=%container.name image=%container.image.repository)"
   priority: WARNING
   tags: [host, container, mitre_execution, mitre_defense_evasion, T1036.005, T1059.004]
-  
+ 
+- list: open_appn_windows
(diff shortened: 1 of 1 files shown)


rules/falco_rules.yaml
```diff
@@ -3397,6 +3397,15 @@
   output: "File execution detected from /dev/shm (proc.cmdline=%proc.cmdline connection=%fd.name user.name=%user.name user.loginuid=%user.loginuid container.id=%container.id evt.type=%evt.type evt.res=%evt.res proc.pid=%proc.pid proc.cwd=%proc.cwd proc.ppid=%proc.ppid proc.pcmdline=%proc.pcmdline proc.sid=%proc.sid proc.exepath=%proc.exepath user.uid=%user.uid user.loginname=%user.loginname group.gid=%group.gid group.name=%group.name container.name=%container.name image=%container.image.repository)"
   priority: WARNING
   tags: [host, container, mitre_execution, mitre_defense_evasion, T1036.005, T1059.004]
+  
+- rule: Get Information About Open Application Windows
+  desc: Detects attempts to get information about open application windows
+  condition: >
```
darryk10, line 3403: As best practices we should add the evt.type at the beginning if possibile so in this case you can add `spawned_process` macro. 
In addition I would move the process names in a proper list so the condition would be easily readable.

darryk10 (changes requested):
Hi @ashutoshmore658 thanks for contributing!
Pls add the signed-off-by in the PR and in your commit. I also dropped some comments for a better rule condition.

rules/falco_rules.yaml
ashutoshmore658: Hi darryk10,
Thank you, for your valuable review.
I have addressed the review and pushed a new commit.

ashutoshmore658:
Can I  create an another branch and raise PR's freshly ?

rules/falco_rules.yaml
```diff
@@ -3397,6 +3397,19 @@
...
   priority: WARNING
   tags: [host, container, mitre_execution, mitre_defense_evasion, T1036.005, T1059.004]
+ 
+- list: open_appln_windows
+  items: ["xwininfo", "wmctrl", "xprop", "xlsclients", "xdotool", "xrestop", "xwininfo", "xdpyinfo", "glxinfo", "xrandr", "xset", "xdriinfo", "xhost", "xev", "xdotool", "xsetroot"]
+   
+- rule: Get Information About Open Application Windows
+  desc: Detects attempts to get information about open application windows
+  condition: >
+    spawned_process and
```
darryk10, line 3407:
```suggestion
    spawned_process and container and
```

darryk10: I would restrict this rule to container only where using those tools is definitely suspicious. Doing this on host level might cause noise. Open to other feedbacks and ideas on this.

rules/falco_rules.yaml
```diff
@@ -3397,6 +3397,19 @@
...
+ 
+- list: open_appln_windows
+  items: ["xwininfo", "wmctrl", "xprop", "xlsclients", "xdotool", "xrestop", "xwininfo", "xdpyinfo", "glxinfo", "xrandr", "xset", "xdriinfo", "xhost", "xev", "xdotool", "xsetroot"]
+   
+- rule: Get Information About Open Application Windows
+  desc: Detects attempts to get information about open application windows
+  condition: >
+    spawned_process and 
+    (proc.name in (open_appln_windows))
+  enabled: true
```
darryk10, line 3409: you don't need to specify this if enable is true. Pls remove this.

darryk10 (changes requested):

rules/falco_rules.yaml
ashutoshmore658: This seems a good decision pushing the changes right away

rules/falco_rules.yaml
```diff
@@ -3397,6 +3397,18 @@
...
   priority: WARNING
   tags: [host, container, mitre_execution, mitre_defense_evasion, T1036.005, T1059.004]
+ 
+- list: open_appln_windows
+  items: ["xwininfo", "wmctrl", "xprop", "xlsclients", "xdotool", "xrestop", "xwininfo", "xdpyinfo", "glxinfo", "xrandr", "xset", "xdriinfo", "xhost", "xev", "xdotool", "xsetroot"]
+   
+- rule: Get Information About Open Application Windows
+  desc: Detects attempts to get information about open application windows
+  condition: >
+    spawned_process and container and
```
darryk10, line 3407: is this rule designed to be just in container or on Host as well? I suggest to modify the Rule Name accordingly. In the condition you added in container so I suggest to add to change the name in " Launch Application Windows in Container". In case you want to apply this rule on Host as well I suggest to remove the container from the condition.

darryk10 (changes requested):

rules/falco_rules.yaml
ashutoshmore658: The rule is designed to identify the activity of getting information about open application windows in a container as well as in Linux host. So I will remove the container from the rule conditions.Pushing the changes right away.

PR #53, opened 2023-05-02
Adding AKS rules

migg24:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/area rules

> /area registry

> /area build

> /area documentation

Adding rules for running Falco on Azure Kubernetes Services (AKS) with less false positives.

Fixes #

I hope this is fine and I did not miss any rules for contributing.

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -1030,7 +1030,7 @@
   tags: [host, container, filesystem, mitre_discovery, T1005]
 
 - list: safe_etc_dirs
-  items: [/etc/cassandra, /etc/ssl/certs/java, /etc/logstash, /etc/nginx/conf.d, /etc/container_environment, /etc/hrmconfig, /etc/fluent/configs.d]
+  items: [/etc/cassandra, /etc/ssl/certs/java, /etc/logstash, /etc/nginx/conf.d, /etc/container_environment, /etc/hrmconfig, /etc/fluent/configs.d. /etc/alertmanager]
 
 - macro: fluentd_writing_conf_files
   condition: (proc.name=start-fluentd and fd.name in (/etc/fluent/fluent.conf, /etc/td-agent/td-agent.conf))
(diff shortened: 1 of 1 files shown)


rules/falco_rules.yaml
```diff
@@ -2467,7 +2468,8 @@
      quay.io/prometheus-operator/prometheus-operator, k8s.gcr.io/ingress-nginx/kube-webhook-certgen,
      registry.k8s.io/ingress-nginx/kube-webhook-certgen, quay.io/spotahome/redis-operator,
      registry.opensource.zalan.do/acid/postgres-operator, registry.opensource.zalan.do/acid/postgres-operator-ui,
-     rabbitmqoperator/cluster-operator)
+     rabbitmqoperator/cluster-operator, quay.io/kubecost1/opencost-ui, quay.io/kubecost1/kubecost-cost-model,
```
darryk10, line 2471: can you provide any documentation to attach re the need to have those images as privileged?
`quay.io/kubecost1/opencost-ui, quay.io/kubecost1/kubecost-cost-model`

darryk10 (changes requested):
Thanks for your PR, seeing changes to reduce the noise is always great!
I dropped just a comment on a couple of images if you could provide further info would be appreciated.

rules/falco_rules.yaml
migg24: Good question, I checked https://www.opencost.io/docs/ but could not find anything. I removed `opencost-ui` again after checking logs again. The reason I added this here is because I see a lot of these in our logs
```
11:50:12.215643425: Notice Unexpected connection to K8s API Server from container (command=app pid=12774 k8s.ns=monitoring k8s.pod=opencost-6bb8f7dc6c-zz94p container=0a5b27ba8701 image=quay.io/kubecost1/kubecost-cost-model:prod-1.101.2 connection=10.1.x.x:44174->10.0.0.1:443)
```

darryk10 (approved):
Thanks @migg24,
LGTM

PR #54, opened 2023-05-11
Add rule for detecting archive and compression activity

0xSh4dy:
Signed-off-by: 0xsh4dy <rakshitawasthi17@gmail.com> ([clouddefense.ai](https://www.clouddefense.ai/))

**What type of PR is this?**

/kind feature

/area rules

Add rule for detecting archive and compression activity

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3398,6 +3398,20 @@
   priority: WARNING
   tags: [host, container, mitre_execution, mitre_defense_evasion, T1036.005, T1059.004]
 
+- list: encryption_proc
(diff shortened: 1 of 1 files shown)


loresuso (commented):
True that archive and compression could be used when exfiltrating data, but how do we discriminate legit actions and non-legit ones? I believe this rule may trigger a lot of false positives.

0xSh4dy:
@loresuso could you please share an example of a false positive that may be triggered. I'll work on it.

loresuso:
I don't have any in mind at the moment, but consider an activity where maybe you compile and package an application in a .tar.gz, that would trigger the rule but could be something that is completely expected

PR #56, opened 2023-05-15
adds rule for detection of Command Execution Via Utility

ashutoshmore658:
Signed-off-by: ashutoshmore658 ashutoshmore658@gmail.com (CloudDefense.AI)

**What type of PR is this?**

> /kind feature

> /area rules

This PR adds a rule for detection of any command execution via utility i.e. command executed not through terminal. As this type of action posses severe security threat, We need to consider this issue.


rules/falco_rules.yaml
```diff
@@ -3397,6 +3397,29 @@
...
   priority: WARNING
   tags: [host, container, mitre_execution, mitre_defense_evasion, T1036.005, T1059.004]
+  
+- list: other_utilities
+  items: [awk, sed, grep, python, perl, bash, cron, systemd-timer, atd]
+  
+- rule: Execute Command Via Utility
+  desc: Detects execution of commands via utilities commonly used to parse text, scripting languages, and system utilities
+  condition: >
+    (evt.type = execve and
```
loresuso, line 3407: There are multiple ways a process may be spawned, and we should take that into account. The macro `spawned_process` should be more accurate.

rules/falco_rules.yaml
```diff
@@ -3397,6 +3397,29 @@
...
+     (proc.name in (other_utilities) or
+      proc.cmdline contains awk or
+      proc.cmdline contains sed or
+      proc.cmdline contains grep or
+      proc.cmdline contains python or
+      proc.cmdline contains perl or
+      proc.cmdline contains bash or
+      proc.cmdline contains cron or
+      proc.cmdline contains systemd-timer or
+      proc.cmdline contains atd))
```
loresuso, line 3417: Here we are triggering this rule every time one of these programs is executed, I don't think it is what we want (also by reading the description of T1202). Maybe you want to detect execution of these programs whenever they are not   a child of most used shells? To do that, check out the `proc.aname` field and the `shell_proc_binaries` macro

PR #67, opened 2023-05-25
update: add extra CI checks for rules changes

jasondellaluce:
**What type of PR is this?**

/kind feature

/area build

This PR serves to experiment options for making our CI stronger.


maxgio92:
Awesome work @jasondellaluce! <3

jasondellaluce:
> Awesome work @jasondellaluce!

Shoutout to @loresuso as well!

jasondellaluce:
Now ready for review! @falcosecurity/rules-maintainers

.github/workflows/ci.yaml
```diff
@@ -20,6 +22,135 @@ jobs:
...
+          ./.github/compare-rule-files.sh \
+              ${{ matrix.rules-file }} \
+              result.txt \
+              build/checker/rules-check \
+              "falcosecurity/falco-no-driver:$FALCO_VERSION"
+          if [ -s result.txt ]; then
+            echo "comment_file=result.txt" >> $GITHUB_OUTPUT
+          fi
+
+      - uses: mshick/add-pr-comment@v2
```
FedeDP, line 152: This is great! However, i am a bit worried if this will work when run on a PR coming from a fork. I had similar issues trying to post a comment on libs repo and had to move to `pull_request_target` trigger that is unsafe: https://github.com/falcosecurity/libs/pull/977

FedeDP: That's it: https://github.com/mshick/add-pr-comment#proxy-for-fork-based-prs

RELEASE.md
```diff
@@ -6,30 +6,46 @@ Official Falcosecurity rules releases are automated using GitHub Actions. Each r
 
 In this repo, each ruleset is a standalone YAML file in the `/rules` directory (e.g. `falco_rules.yaml`, `application_rules.yaml`, ...). Each ruleset is released and versioned individually. When we release a ruleset, we do the following process:
 
-1. Determine a new version for the given ruleset (see the [section below](#versioning-a-ruleset))
-2. Create a new Git tag with the name convention `*name*-rules-*version*` (e.g. `falco-rules-0.1.0`, `application-rules-0.1.0`, ...). The naming convention is required due to this repository being a [monorepo](https://en.wikipedia.org/wiki/Monorepo) and in order to be machine-readable.
-3. A GitHub action will validate the repository [registry](./registry.yaml) and release the new OCI artifact in the packages of this repository
+1. Make sure that the `./github/FALCO_VERSIONS` file contains the most recent versions of Falco compatible with the given ruleset. The `master` version must not be used when releasing a ruleset.
```
leogr, line 9: Are we implicitly saying releases can only happen on a release branch? :thinking:

RELEASE.md
```diff
@@ -6,30 +6,46 @@ Official Falcosecurity rules releases are automated using GitHub Actions. Each r
...
 
 ## Versioning a ruleset
 
 The version of the official Falco rulesets is compatible with [SemVer](https://semver.org/) and must be meaningful towards the changes in its structure and contents. To define a new version `x.y.z` for a given ruleset, consider the following guidelines. 
 
+Our automation will detect most of the following criteria and suggest a summary with all the changes relative to each of the three versioning categories (patch, minor, major). This provides a changelog and valuable suggestion on the next version to be assigned to each ruleset. However, be mindful that the versioning process cannot totally automated and always requires human attention (e.g. we can't automatically detect subtle semantic changes in rules). The automated jobs will use the versions of Falco defined in `./github/FALCO_VERSIONS` for validating and checking rulesets. The versions must be line-separated and ordered from the most recent to the least recent. The `master` keyword is valid and indicates the most recent dev version of Falco built from mainline, and must not be used when releasing a ruleset.
```
leogr, line 18: Is `master` just a special case or do we support any valid git ref?

RELEASE.md
```diff
@@ -6,30 +6,46 @@ Official Falcosecurity rules releases are automated using GitHub Actions. Each r
...
+    - Adding or removing exceptions for one or more Falco rules (without increasing `required_engine_version`)
+    - Changing the condition for one or more rules by still preserving their logical security scope (e.g. making them less noisy, matching more events than before)
 - `y` _(minor number)_ is incremented when you add functionality in a backward-compatible manner. In order to be accepted, the new ruleset may mandate updating the version of Falco, changing its configuration, or updating/installing one or more plugins. Examples:
-    - Incrementing the `required_engine_version` number
+    - Incrementing `required_engine_version`
     - Incrementing the `required_plugin_versions` version requirement for one or more plugin
-    - Adding a new plugin version requirement in `required_plugin_versions`
+    - Adding a new plugin version requirement (or an alternative) in `required_plugin_versions`
     - Adding one or more lists, macros, or rules
+    - Adapting the ruleset to a Falco engine version introducing *backward-compatible* changes in the expected ruleset language definitions or file format
```
leogr, line 39: This is not clear to me. Do you mean without changing the `required_engine_version`? Could you provide an example please?

leogr (commented):
Overall, looks great! 

Just a few questions. Please take a look at my comments below.

RELEASE.md
jasondellaluce: Nope. `master` is just a valid Falco version keyword, that can be used to run the latest dev version of Falco in the CI. It has nothing to do with how we manage the ruleset releases.

The message here is that `master` can be used when we need to leverage a not-yet-release version of Falco to run the CI checks for any given reason (maybe a new feature). However, when a ruleset gets tagged it would be preferable to point to a stable released version of Falco. This could apply if either we decide to have branch releases or not in the future.

RELEASE.md
jasondellaluce: It's a special case. Under the hood, that's the tag of the container image used for running Falco. As such, I think it's ok to document that the only safe cases available are `master` and the official falco version tags (e.g. `0.34.1`).

RELEASE.md
jasondellaluce: > Adapting the ruleset to a Falco engine version introducing *backward-compatible* changes in the expected ruleset language definitions or file format

For example, where in the past we introduced the `append` or `exceptions` features in the YAML format of Falco rules, I would have expected only the minor version of the ruleset to be bumped. This is because migrating to the new feature didn't require changes in previously-existing rulesets if not simply "adding" new things inside it. This would have not broken expectations about all the previosly-defined lists, macros, rules, and how they are resolved by the engine.

> Adapting the ruleset to a Falco engine version introducing *backward-incompatible* changes in the expected ruleset language definitions or file format

Think about the case in which we fix some bugs (such as making the string escaping parsing sound), or that we alter the way the Falco engine resolves rules and their dependencies. Indeed, this would be a breaking change from the standpoint of the Falco engine, but it will also be a breaking change for the ruleset as well (thus requiring bumping the major version). For example, this could break assumptions about when a rule triggers or doesn't, what could be the expected alert, or potentially can change the `append` semantics on which users may rely on when appending their definitions to the default rules.

RELEASE.md
```diff
@@ -6,30 +6,46 @@ Official Falcosecurity rules releases are automated using GitHub Actions. Each r
...
+    - Increasing the priority of a rule
+    - Changing the output fields of a rule (without increasing `required_engine_version`)
+    - Adding or removing exceptions for one or more Falco rules (without increasing `required_engine_version`)
+    - Changing the condition for one or more rules by still preserving their logical security scope (e.g. making them less noisy, matching more events than before)
 - `y` _(minor number)_ is incremented when you add functionality in a backward-compatible manner. In order to be accepted, the new ruleset may mandate updating the version of Falco, changing its configuration, or updating/installing one or more plugins. Examples:
-    - Incrementing the `required_engine_version` number
+    - Incrementing `required_engine_version`
     - Incrementing the `required_plugin_versions` version requirement for one or more plugin
-    - Adding a new plugin version requirement in `required_plugin_versions`
+    - Adding a new plugin version requirement (or an alternative) in `required_plugin_versions`
```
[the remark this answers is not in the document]
jasondellaluce: > Adding a plugin version requirement alternative in required_plugin_versions

This refers to when an alternative is added to an already-existing plugin dependency. For example when we updated:
```yaml
- required_plugin_versions:
  - name: k8saudit
     version: x.x.x
```
To:
```yaml
- required_plugin_versions:
  - name: k8saudit
     version: x.x.x
     alternatives:
       - name: k8saudit-eks
          version: y.y.y
```

> Adding a new plugin version requirement (or an alternative) in required_plugin_versions

This refers to when we add a new plugin dependency that didn't yet exist in the ruleset. The new dependency can either have alternatives right from the beginning or not. For example, if the `k8saudit` dependency didn't exist in a given ruleset, adding it with either no alternative (first yaml above) or with an alternative (second yaml above), would still bump the minor due to the dependency not being existing before.

Maybe a good reworking may be *Adding a new plugin version requirement (with or without alternatives) in required_plugin_versions*, WDYT?

RELEASE.md
leogr: Ok, so also, `latest` and `0.35.0-alpha5 should work, basically any [published tag](https://hub.docker.com/r/falcosecurity/falco/tags), shouldn't they? 

It's worth documenting that, IMO.

RELEASE.md
leogr: > > Adapting the ruleset to a Falco engine version introducing _backward-compatible_ changes in the expected ruleset language definitions or file format
> 
> For example, where in the past we introduced the `append` or `exceptions` features in the YAML format of Falco rules, I would have expected only the minor version of the ruleset to be bumped. This is because migrating to the new feature didn't require changes in previously-existing rulesets if not simply "adding" new things inside it. This would have not broken expectations about all the previosly-defined lists, macros, rules, and how they are resolved by the engine.

This makes sense if we don't bump `required_engine_version` in such a case.

Rationale: since there's no way to know from `required_engine_version` if a backward-incompatible change has been introduced, we can only assume that the change might not be backward-compatible.

(it would have been different if `required_engine_version` was a semver-like version number. Are we assuming it will be? :thinking: )

jasondellaluce: > (it would have been different if required_engine_version was a semver-like version number. Are we assuming it will be? 🤔 )

I assume it *may* be in the future. I don't want to exclude this possibility. Let me try to rephrase it better.

RELEASE.md
jasondellaluce: Yep

leogr (approved):
Looks very good to me know.

Great job, thank you :pray:

PR #72, opened 2023-06-01
fix(ci): skip rules jobs when no rule is changed

jasondellaluce:
**What type of PR is this?**

/kind bug

/area build

Testing the CI, and fixing it to skip rules jobs when no ruleset is changed.


jasondellaluce:
This should be ready to go!

Andreagit97 (approved):
/approve

PR #76, opened 2023-06-07
new(proposals): rules adoption, management and maturity framework

incertum:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

/kind documentation

> /kind failing-test

/kind feature

> /area rules

> /area registry

> /area build

/area documentation

Formalize rules adoption, management and maturity framework.
@falcosecurity/core-maintainers @falcosecurity/rules-maintainers 

Fixes #


proposals/20230605-rules-adoption-management-maturity-framework.md
```diff
@@ -0,0 +1,222 @@
...
+
+> Falco introduces a new key called "maturity" to each rule with levels including "Stable", "Incubating", "Sandbox" and "Deprecated" to guide the adoption process. This key reflects the robustness and relevance of each predefined rule in the [falcosecurity/rules](https://github.com/falcosecurity/rules/blob/main/rules/) repository. Each rule must go through the "Incubating" state before advancing to "Stable" as a minimum requirement. Only "Stable" rules will be enabled by default. 
+
+The maturity level of the rules, however, does not directly reflect their potential for generating noise in the adopters' environment. This is due to the unique and constantly changing nature of each environment, especially in cloud environments, making it challenging to accurately predict the impact of rules.
+
+Newcomers to Falco will be encouraged to start by configuring their setup with introductory rules labeled as "Falco's default rules" (maturity: "Stable"). These rules will be based on syscall and container events and will live in the established [falco_rules.yaml](https://github.com/falcosecurity/rules/blob/main/rules/falco_rules.yaml).
```
maxgio92, line 56: How we could / would want to distinguish the maturity level (e.g. `stable`) from the importance / probability that those rules are as agnostic as possible and likely useful for a large number of environments?
For example we can have stable Falco rules, and stable but environment-specific rules.

Is it enough to split Falco rules from application rules?

proposals/20230605-rules-adoption-management-maturity-framework.md
```diff
@@ -0,0 +1,222 @@
...
+
+
+### Guidance on Rule Customization and Tuning
+
+Each rule tagged as "Stable" will provide clear guidance on how it can be tuned, customized, or combined with other rules if applicable. Over time, a catalog of general best tuning practices will be developed. For specific examples, please refer to Appendix 2. 
+
+
+### Setting Expectations for Adopters
+
+Setting clear expectations for adopters of Falco is crucial. While Falco is a powerful security monitoring tool, it's important to be aware of its limitations and maintain realistic expectations. Please refer to Appendix 3 for specific examples of current limitations of Falco.
```
maxgio92, line 107: What do you think if we also guide the community to lateral tools for specific use cases Falco does not cover?
For example, in case the community user would need projects for threat detection with network analysis, could be guided to projects like Suricata.

maxgio92:
Thank you @incertum, great work!
And I like it. Just left some side questions :)

proposals/20230605-rules-adoption-management-maturity-framework.md
incertum: Happy to discuss, for me from my perspective "Stable" shall not indicate stable from a software engineering perspective, but rather tags rules that we recommend any adopter regardless of environment or industry shall adopt. The "Stable" rules aim to detect indicators of compromise of top cyber threats relevant to any environment.

Re your second point I actually would love to understand what "application" means for everyone, because I am confused by for example the existing application rules https://github.com/falcosecurity/rules/blob/main/rules/application_rules.yaml as they are all based on network related system calls and the "application" here comes from the port definitions. An alternative way of looking at application could be to really only focus on higher up in the stack such as auth bypasses ...

Perhaps we should clearly define "system" and "application" level detections for Falco? Another example are detections around Kube audit logs, they are more "control plane" ... in summary clear definitions we all agree upon would be amazing. @maxgio92 would you have a suggestion for those definitions?

proposals/20230605-rules-adoption-management-maturity-framework.md
incertum: This would be great! If I recall correctly, we wanted to add some more generic educational resources on the website as well, also around eBPF or any topic / domain falco touches.

proposals/20230605-rules-adoption-management-maturity-framework.md
```diff
@@ -0,0 +1,222 @@
...
+
+This guide will also include information on how the community is intended to share rules. As a general principle, the project's main objective is to curate an optimal set of "Stable" rules that are included in the main [falco_rules.yaml](https://github.com/falcosecurity/rules/blob/main/rules/falco_rules.yaml) file and also published as OCI artifacts. However, there may be opportunities to create additional platforms for ad-hoc rules sharing that do not necessarily need to meet the strict criteria set by the Falco maintainers.
+
+
+The minimum criteria for rules creation shall include:
+
+*Correctness*
+
+The rule must be both syntactically and grammatically correct and should evaluate to true during successful end-to-end tests. Furthermore, it needs to accurately detect the intended cyber threats, specifically the Tactics, Techniques, and Procedures (TTPs).
```
IceManGreen, line 76: @incertum I am writting down some things here that we discussed about Mitre to trace it.
So beyond the correctness of the rules regarding the proper Falco syntax, I would propose more precisions about the correctness of Falco rules regarding standards (or de facto standards), like the Mitre ATT&CK, when the rules are using extra tags.

The rules already use the naming system `mitre_<mitre phase name>`, so maybe we can stick with it to detect when a rule is using a Mitre extra tag, or simply a technique id like `T1021.004`.

I am still pretty confident with STIX2 standard to fetch the Mitre ATT&CK data, so maybe it can appear in this section ?
Something like :

"The rule must be both syntactically and grammatically correct and should evaluate to true during successful end-to-end tests. 
Furthermore, it needs to accurately detect the intended cyber threats, specifically the Tactics, Techniques, and Procedures (TTPs), _against the Mitre ATT&CK framework.
Both the community and the Falco experts will benefit from the falco mitre checker module to audit the default rules or custom rules against the STIX2 data from Mitre CTI._ "

proposals/20230605-rules-adoption-management-maturity-framework.md
maxgio92: Thanks @incertum, I got the meaning of stable here, so it comprises also the relevance.
What creates confusion a bit is the fact that I'd expect the stability levels (`stable`, `incubating`, `sandbox`) would be transient - e.g. incubating rules might become stable, sandbox ones incubating.
If I undestand correctly, instead, these levels are not transient - e.g. incubating rules whether correct or not, are supposed to be relevant only for a subset of environment. Am I wrong?

I agree with the fact that we should clarify the distinction between [application](https://github.com/falcosecurity/rules/blob/main/rules/application_rules.yaml) and [system / core](https://github.com/falcosecurity/rules/blob/main/rules/falco_rules.yaml) rules.
From my understanding, application ones are what are not i.e. stable, in the sense of not being rules that should detect threats relevant to any environment, but instead, threats relevant to specific environments, and specifically related to optional user space applications, in some way.
I don't think, as far as I undestand, application refers to the stack level, in the OSI sense.

> Then, additional data sources, through plugins, can relate to additional application rules (like Kubernetes Audit rules with related plugin), but application rules might not necessary need plugins.

In any case I think we should clarify the meaning of application here. And I'd like to map this distinction (system/core and application) to the stability levels (stable, incubating, sandbox, deprecated).
If the above is correct, I think it could make sense and make the vocabulary more consistent. What do you think?

incertum: @maxgio92 I agree, using the status levels felt appealing, but maybe they not quite right yet. Good news is I think we appear to align on what we are looking for, just need to find a good leveling that is expressive, self-explanatory and creates the least confusion - with that do you have ideas? 🙃 

I like your thinking wrt to defining "application" a lot actually! So in summary what would you suggest we change it to? I think we are close, just need  maybe 2-3 more rounds of editing.

proposals/20230605-rules-adoption-management-maturity-framework.md
incertum: Thank you a lot @IceManGreen I'll work it in and leave it more general for now and then we decide on the technical details for the Falco Mitre checker module you are looking to create in your proposal ❤️ !

proposals/20230605-rules-adoption-management-maturity-framework.md
maxgio92: I propose to mimic the current repository [scope](https://github.com/falcosecurity/evolution/blob/main/REPOSITORIES.md#scope) **and** [status](https://github.com/falcosecurity/evolution/blob/main/REPOSITORIES.md#status) parameters.

Scopes: _core_ (e.g. [core rules](https://github.com/falcosecurity/rules/blob/main/rules/falco_rules.yaml)) and _ecosystem_ (e.g. [application rules](https://github.com/falcosecurity/rules/blob/main/rules/application_rules.yaml) and matching the *application* meaning which applies to application rules), that apply to rule set and refers to the relevance of the rules for the community use cases.

And for each rule set, orthogonal to the scope, the status maturity level, that represents the maturity stage of the rules (_stable_, _incubating_, _sandbox_, _deprecated_), of which incubating and sandbox ones likely try to converge to the stable level.

I'd like to have @leogr's feedback too, who worked on the parallel parameters for the repositories, and sure that can add important feedback :-)

incertum: This is great, in summary 2 new keys `maturity` and `scope`. Would try to adjust the wordings in a way that still conveys that transitioning to "stable" maturity has to provide some broader value than string matching on fields in a particular context that could be easily circumvented. Would this resonate with all of you?

Perhaps the system vs application discussion is another discussion we should continue, but it does not have to block this framework. Could see a solution where we just have all rules that are *not* plugin rules in the same file for easier maintenance. Distinctions shall be based on the new keys and possible future extensions of the tags key?

Yes @leogr plz share your perspective, ty!

proposals/20230605-rules-adoption-management-maturity-framework.md
```diff
@@ -0,0 +1,222 @@
...
+- [Tutorials](https://falco.org/docs/tutorials/) and [blog posts](https://falco.org/tags/rules/) related to rules are available on the Falco website, along with numerous references to webinars and conference talks.
+- Falco features a CLI tool called [falcoctl](https://github.com/falcosecurity/falcoctl) to manage the lifecycle of rules (installation, updates), see [blog post](https://falco.org/blog/falcoctl-install-manage-rules-plugins/).
+- The rules framework of Falco is battle-tested and proven to be reliable in production environments. It is used by a diverse range of [organizations](https://github.com/falcosecurity/falco/blob/master/ADOPTERS.md).
+- Existing rules yaml files
+  - Main Falco rules ([falco_rules.yaml](https://github.com/falcosecurity/rules/blob/main/rules/falco_rules.yaml)) based on syscall and container events.
+  - [application_rules.yaml](https://github.com/falcosecurity/rules/blob/main/rules/application_rules.yaml) also contains rules based on network-related syscalls, which may seem misleading if you expect those rules to be present in the regular "falco_rules.yaml" file.
```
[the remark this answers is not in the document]
incertum: Thanks @TheFoxAtWork -- @falcosecurity/rules-maintainers can we merge those rules into the existing falco_rules.yaml? What needs to happen to facilitate this transition?

proposals/20230605-rules-adoption-management-maturity-framework.md
incertum: I have made some adjustments, PTAL, thanks.

incertum:
@TheFoxAtWork thanks for publishing "some areas of focus for the project to undertake prior to Graduation Public Comment opening" https://github.com/cncf/toc/pull/956#issuecomment-1599367728.

We are taking ownership of addressing acknowledged gaps regarding rules adoption and maturity leveling. We would greatly appreciate any additional suggestions for improvement and early feedback to support us.

proposals/20230605-rules-adoption-management-maturity-framework.md
LucaGuerra: Thank you both for the discussion.

I appreciate the idea of trying to make it clearer to users and contributors the purpose of each file we distribute. I took a look at the `maturity` and `scope` paragraphs and the difference wasn't completely clear to me. I would propose we start with one level, `maturity` (stable, incubating, sandbox), that we can also use to distribute the rules (@leogr and @incertum had a great idea about the distribution part). This would allow us to move forward, be clearer so people don't have to figure out the differences between the two parts, make the repo easier to maintain and contribute to. If we understand that we need another level we can add it at a future iteration.

proposals/20230605-rules-adoption-management-maturity-framework.md
```diff
@@ -0,0 +1,244 @@
...
+
+This guide will also include information on how the community is intended to share rules. As a general principle, the project's main objective is to curate an optimal set of "Stable" rules that are included in the main [falco_rules.yaml](https://github.com/falcosecurity/rules/blob/main/rules/falco_rules.yaml) file and also published as OCI artifacts. However, there may be opportunities to create additional platforms for ad-hoc rules sharing that do not necessarily need to meet the strict criteria set by the Falco maintainers.
+
+
+The minimum criteria for rules creation shall include:
+
+*Correctness*
+
+The rule must be both syntactically and grammatically correct and should evaluate to true during successful end-to-end tests. Furthermore, it needs to accurately detect the intended cyber threats, specifically the Tactics, Techniques, and Procedures (TTPs) aligned with the [Mitre ATT&CK](https://attack.mitre.org/) framework. To ensure this, a new [Mitre Checker Module](https://github.com/falcosecurity/rules/issues/88) will be utilized to audit the correct Mitre tags and references.
```
LucaGuerra, line 98: This sounds a bit confusing but I agree with the sentiment.

Big +1 on:
* correctness with the language
* consistency with the name/description
* if we have any kind of test, they need to pass

On the other hand, it mixes it with availability of e2e tests and a potential tool to verify rules against a specific compliance framework. While we can do many things with our rules and the more automated tests we have the better it is, mentioning a potential tool that works with a specific framework sounds a bit out of scope for this definition.

incertum: Agreed, this seems out of scope for this proposal and we can include these items in the style guide.

proposals/20230605-rules-adoption-management-maturity-framework.md
incertum: Considering all feedback I have pushed a new version. I like the callout of  keeping things more simple in the beginning.

incertum (commented):
I have an additional question:

Each rule will be augmented by a new yaml key called `maturity`. We have also discussed the possibility of splitting rules into various files:

```
falco_rules.yam
falco_rules.incubating.yaml
falco_rules.sandbox.yaml
```

I can see multiple benefits in doing so, while recognizing that it can be disruptive to some adopters. Therefore, I would also be ok to first categorize the rules in the existing rules file and subsequently for Falco 0.37 split the rules into separate files to allow for some time to communicate the upcoming changes to adopters and perhaps in that way we also have more time to improve `falcoctl`?

leogr (commented):
Firstly, I want to thank you for your fantastic work on this proposal :star_struck: :pray:  

I'm generally pleased with it and only have a couple of suggestions I tried to describe in detail for clarity :point_down: 

**Dividing rules into separate files (based on maturity)**
**TL;DR** _I agree with the proposed method of splitting rule files._

> Each rule will be augmented by a new yaml key called `maturity`. We have also discussed the possibility of splitting rules into various files:
> 
> ```
> falco_rules.yam
> falco_rules.incubating.yaml
> falco_rules.sandbox.yaml
> ```
> 
> I can see multiple benefits in doing so, while recognizing that it can be disruptive to some adopters. Therefore, I would also be ok to first categorize the rules in the existing rules file and subsequently for Falco 0.37 split the rules into separate files to allow for some time to communicate the upcoming changes to adopters and perhaps in that way we also have more time to improve `falcoctl`?

I agree with breaking down rules into separate files, given our artifact distribution system we have in place specifically designed for such purposes. Moreover, only `falco_rules.yaml` should be part of the Falco installation package. Users desiring to use other rule files can install them via `falcoctl`.

I'm also fine with dividing this process into two stages (i.e., merely categorizing in 0.36 and splitting files in 0.37). Nevertheless, I suggest we make this decision later once we clearly understand the impact. As noted during yesterday's maintainer call, we should contemplate announcing our changes in August. Users will also get an opportunity to test these changes in the dev builds we intend to release by the end of this month. If any complaints arise, we should endeavor to resolve the issues and improve usability.

A side note: as also discussed yesterday, `application_rules.yaml` should remain a separate file and be downgraded to `sandbox` or potentially `deprecated` status. This will not cause any disruption for users, as Falco has never defaulted to loading this file, and all rules within it are commented on (they were originally intended as examples and not as ready-to-use rules).


**Utilizing tags for maturity levels**
**TL;DR** _I'm in favor of using tags, as they're well-established and don't introduce a new component that partially duplicates an existing one._

I advise against introducing a new yaml key, as Falco already possesses a reliable mechanism for enabling/disabling rules via tags:

```
  -T <tag>                      Disable any rules with a tag=<tag>. This option can be passed multiple times. Cannot be mixed with -t
  -t <tag>                      Only run those rules with a tag=<tag>. This option can be passed multiple times. Cannot be mixed with 
                                -T/-D.
```

Thus, as a counter-proposal, I propose we define the following tags and make use of the `-T/-t` feature:

- `maturity_stable`
- `maturity_incubating`
- `maturity_sandbox`
- `maturity_deprecated`

Eventually, we can also add these selectors to `falco.yaml`, thereby providing more transparent configurability through the configuration file, as outlined in your proposal.

proposals/20230605-rules-adoption-management-maturity-framework.md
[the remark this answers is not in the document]
leogr: As discussed yesterday, `application_rules.yaml` was never meant to be part of the default ruleset (it just provides some examples) and should not be merged into `falco_rules.yaml`. I reported full details in my [review below](https://github.com/falcosecurity/rules/pull/76#pullrequestreview-1518592957) :point_down: 

A side note: as also discussed yesterday, `application_rules.yaml` should remain a distinct file and be downgraded to `sandbox` or potentially `deprecated` status. This will not cause any disruption for users, as Falco has never defaulted to loading this file, and all rules within it are commented (they were originally intended as examples and not as ready-to-use rules).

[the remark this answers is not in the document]
incertum: ACK, would propose to gather some community feedback around this file and the rules it contains. In addition, looking back to the discussion @maxgio92 and I had we may wanna keep track of the "what are these application rules for Falco vs the other rules?" discussion in a dedicated issue. We should not block this proposal because of this.

incertum:
@leogr feedback incorporated also stating to go with the safer approach first, that is, focus on tagging rules and creating all guides. Gather community feedback early on as you suggested while prioritizing some other improvements in parallel that would make multiple upstream rules files easier to use, adopt and maintain.

leogr:
> @leogr feedback incorporated also stating to go with the safer approach first, that is, focus on tagging rules and creating all guides. Gather community feedback early on as you suggested while prioritizing some other improvements in parallel that would make multiple upstream rules files easier to use, adopt and maintain.

@incertum Thank you!

Is this still `wip:` ? Or are we ready for the final review?

leogr (approved):
LGTM! :pray:

PR #82, opened 2023-06-10
update(rules): add rules for PKT mining pools

blampe:
/kind feature
/area rules

[PacketCrypt](https://docs.pkt.cash/en/latest/mining/) (PKT) is a bandwidth-hard proof-of-work cryptocurrency. This particular coin makes mining viable in environments with limited compute resources but healthy network pipes.

PKT mining requires announcing to a pool, and there are only a few pools available because they need to be the size of a small datacenter to accommodate all of the incoming bandwidth from the protocol.

This PR adds all 4 of the [publicly known](https://docs.pkt.cash/en/latest/mining/#choosing-pools-to-mine-in) PKT pools as of 6/2023 to the `miner_domain` rules.

I've alphabetized `https_miner_domains` but can revert that if desired.

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -2856,6 +2856,7 @@
 
 - list: https_miner_domains
   items: [
+    "btm.antpool.com",
     "ca.minexmr.com",
     "cn.stratum.slushpool.com",
     "de.minexmr.com",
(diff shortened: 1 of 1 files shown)


loresuso (commented):
Hello @blampe and thank you for this PR! 
To let these changes in, we need commit messages to adhere to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), could you please modify it? Thanks

blampe:
> Hello @blampe and thank you for this PR!
> To let these changes in, we need commit messages to adhere to [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), could you please modify it? Thanks

@loresuso sure! Prefixed with `update(rules):` LMK if `add(rules)` or something else is more appropriate!

loresuso:
Thanks @blampe!
Some other PRs were merged, if you rebase on master we can let this in :)

darryk10:
Hi @blampe, does it still make sense to split those domains in http and https? those domains needs to be picked up by falco anyway.

blampe:
> Thanks @blampe!
> Some other PRs were merged, if you rebase on master we can let this in :)

@loresuso apologies, I've been traveling for a couple weeks. This has been rebased on main.

> Hi @blampe, does it still make sense to split those domains in http and https? those domains needs to be picked up by falco anyway. Having specific protocol related lists make them more complicated to maintain. Whatever protocol they use, they need to be detected. Perhaps adding the port 80, 443 in miner_ports would simplify the overall condition and make it easier and more efficient.

@darryk10 that's a good point. I only added them to the respective http/https lists because there seemed to already be a convention. I don't know what the long-term plans are for those lists and I'm happy to do whatever's most maintainable and aligned with the project's long-term goals. I'm not familiar with the behavior of `miner_ports`, and my only hesitation is if that might create false positives due to 80/443 being so ubiquitous. (A rule that flagged traffic to these ports would catch non-crypto traffic, for example.)

PR #83, opened 2023-06-12
fix(rules/falco_rules): solve shadowing issues with "Drop and execute new binary in container"

jasondellaluce:
**What type of PR is this?**

/kind bug

/area rules

The rule `Drop and execute new binary in container` has been introduced in #20 (part of the [`falco-rules-1.0.0` release](https://github.com/falcosecurity/rules/releases/tag/falco-rules-1.0.0)) and has a condition that is very similar to other rules, but that however is more narrow. Being defined before the rules below, the probability of "shadowing" them is quite high (a.k.a. case in Falco only reports the first rule matching, thus not triggering rules defined after in the YAML order):

- `Terminal shell in container`
- `Launch Package Management Process in Container`
- `Netcat Remote Code Execution in Container`
- `Launch Suspicious Network Tool in Container`
- `Launch Remote File Copy Tools in Container`
- `The docker client is executed in a container`
- `Container Run as Root User`
- `Debugfs Launched in Privileged Container`
- `Mount Launched in Privileged Container`
- `Launch Ingress Remote File Copy Tools in Container`

My proposal is to move `Drop and execute new binary in container` down in the declaration order. The rule has lots of value, but can't shadow other important rules.

cc @incertum @loresuso 

Fixes #

On the long term, we could consider making Falco trigger all rules instead of the first one, but that would be a new feature for a future Falco release, and not a patch for the existing rulesets.

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -496,30 +496,6 @@
     WARNING
   tags: [host, container, filesystem, mitre_discovery, T1546.004]
 
-# List of allowed container images that are known to execute binaries not part of their base image.
-# Users can use this list to better tune the rule below (i.e reducing false positives) by considering their workloads, 
-# since this requires application specific knowledge.
-- list: known_drop_and_execute_containers
-  items: []
-
-- rule: Drop and execute new binary in container
-  desc:
-    Detect if an executable not belonging to the base image of a container is being executed.
-    The drop and execute pattern can be observed very often after an attacker gained an initial foothold.
-    is_exe_upper_layer filter field only applies for container runtimes that use overlayfs as union mount filesystem.
-  condition: >
-    spawned_process
-    and container
-    and proc.is_exe_upper_layer=true 
-    and not container.image.repository in (known_drop_and_execute_containers)
-  output: > 
-    Executing binary not part of base image (user=%user.name user_loginuid=%user.loginuid user_uid=%user.uid comm=%proc.cmdline exe=%proc.exe container_id=%container.id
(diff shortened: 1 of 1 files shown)


incertum:
@jasondellaluce ty! I think the "shadowing" is something we should discuss more in detail in the new rules maturity framework as we build it out. Perhaps @jasondellaluce would you have ideas on how to expose something easy to use for end users to dictate the ordering? For example, I have a custom Go patching script that re-orders based on a predefined priority list which includes custom rules as well.

But this is beyond this PR, LGTM!

incertum:
One more thought, should we add more output fields to the other rules that could be a new executable?

```
evt.arg.flags
proc.exe_ino.ctime_duration_proc_start
```

jasondellaluce:
That sounds like a good idea to me. @loresuso word to you on this.

loresuso:
I agree that in the long term, Falco should trigger each rule in which the condition is satisfied. I think we should start prioritizing this, since the shadowing problem should be solved from its root cause. 

I am ok with moving the rule down in the file and as Melissa was saying, I think we can print out the `evt.arg.flag` in all the other rules so that we can understand if `exe_upper_layer` was also set from the output.

>For the future, rule/macro/list reordering is something we should catch in automatic change checks. If we can't predict the major/minor/nature of the reordering, my opinion would be to add an "other/extra changes" section for things like this.

agreed!

jasondellaluce:
Is everyone onboard with the current change? I would like to issue a v1.0.1 patch release of the Falco rules including this.

incertum:
Yes happy to approve once I can ...

incertum (approved):
/approve

pmusa:
Hey folks, thank you for the PR. How can I know when is this going to be GA? This will break the labs for the Falco 101 workshop, and we have 10+ deliveries in the next 2 weeks.

PR #85, opened 2023-06-13
Exploit Detection for CVE-2023-33246 - RocketMQ

Loginsoft-Research:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

 /area rules

> /area registry

> /area build

> /area documentation

Added detection for exploit activity and exploit attempt of RocketMQ unauthenticated arbitrary file write vulnerability which leads to remote code execution (CVE-2023-33246)

POC Reference:
https://github.com/Malayke/CVE-2023-33246_RocketMQ_RCE_EXPLOIT/blob/main/check.py
https://github.com/hheeyywweellccoommee/CVE-2023-33246-rnkku

Fixes #

Looking forward to feedback and suggestions
[cve_2023_33246_output.txt](https://github.com/falcosecurity/rules/files/11738207/cve_2023_33246_output.txt)
.

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3426,23 +3426,20 @@
   condition: (proc.name contains NettyServer and proc.env contains ROCKETMQ)
 
 - macro: rocketmq_check_payload_for_conf_update
-  condition: (evt.type=read and evt.buffer contains 'code":25' and evt.buffer contains serializeTypeCurrentRPC and evt.buffer contains JSON)
+  condition: ((fd.typechar='4' or fd.typechar='6') and evt.type=read and evt.buffer contains 'code":25' and evt.buffer contains serializeTypeCurrentRPC and evt.buffer contains JSON)
 
 - rule: Apache Rocketmq CVE-2023-33246 Exploit Detection
   desc: >
     CVE-2023-33246 is exploited at the function used to update configuration & can be detected by looking at the buffer for broker code-25 and serializeTypeCurrentRPC function.
   condition: >
-    rocketmq_check_netty_server  and rocketmq_check_payload_for_conf_update
+    rocketmq_check_netty_server and rocketmq_check_payload_for_conf_update and evt.dir=<
   output: exploit detected CVE-2023-33246 (event=%evt.type server_ip=%fd.sip server_port=%fd.sport proto=%fd.l4proto parent=%proc.pname process=%proc.name command=%proc.cmdline container_id=%container.id)
   priority: CRITICAL
-  tags: [host, CVE_2023_33246, rocketmq, apache,Mitre_Initial_Access, Mitre_Execution,TA0001, TA0002, T1133, T1059]
+  tags: [host, container, CVE_2023_33246, rocketmq, apache,Mitre_Initial_Access, Mitre_Execution,TA0001, TA0002, T1133, T1059]
 
 
 - macro: rocketmq_check_env_variable
   condition: (proc.env contains ROCKETMQ)
-
-- macro: check_file_openat
-  condition: (fd.type=file and evt.type=openat)
 
 - macro: rocketmq_check_properties_file_write_conf
   condition: (fd.nameraw contains conf and fd.nameraw endswith '.properties')
@@ -3451,7 +3448,7 @@
   desc: >
     Look for the file ends with '.properties' created under folder 'CONF' upon exploit attempt
   condition: >
-    rocketmq_check_env_variable and check_file_openat and rocketmq_check_properties_file_write_conf
+    rocketmq_check_env_variable and open_write and rocketmq_check_properties_file_write_conf and evt.dir=<
   output: CVE-2023-33246 exploit attempt detected (event=%evt.type filepath=%fd.name server_ip=%fd.sip server_port=%fd.sport proto=%fd.l4proto parent=%proc.pname process=%proc.name command=%proc.cmdline container_id=%container.id)
   priority: WARNING
-  tags: [host, CVE_2023_33246, rocketmq, apache,Mitre_Initial_Access, Mitre_Execution,TA0001, TA0002, T1133, T1059]
+  tags: [host, container, CVE_2023_33246, rocketmq, apache,Mitre_Initial_Access, Mitre_Execution,TA0001, TA0002, T1133, T1059]
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3420,3 +3420,38 @@
 # Application rules have moved to application_rules.yaml. Please look
 # there if you want to enable them by adding to
 # falco_rules.local.yaml.
+
+
+- macro: rocketmq_check_netty_server
+  condition: (proc.name contains NettyServer and proc.env contains ROCKETMQ)
+
+- macro: rocketmq_check_payload_for_conf_update
+  condition: (evt.type=read and evt.buffer contains 'code":25' and evt.buffer contains serializeTypeCurrentRPC and evt.buffer contains JSON)
+
+- rule: Apache Rocketmq CVE-2023-33246 Exploit Detection
+  desc: >
+    CVE-2023-33246 is exploited at the function used to update configuration & can be detected by looking at the buffer for broker code-25 and serializeTypeCurrentRPC function.
+  condition: >
+    rocketmq_check_netty_server  and rocketmq_check_payload_for_conf_update
+  output: exploit detected CVE-2023-33246 (event=%evt.type server_ip=%fd.sip server_port=%fd.sport proto=%fd.l4proto parent=%proc.pname process=%proc.name command=%proc.cmdline container_id=%container.id)
+  priority: CRITICAL
+  tags: [host, CVE_2023_33246, rocketmq, apache,Mitre_Initial_Access, Mitre_Execution,TA0001, TA0002, T1133, T1059]
+
+
+- macro: rocketmq_check_env_variable
+  condition: (proc.env contains ROCKETMQ)
+
+- macro: check_file_openat
+  condition: (fd.type=file and evt.type=openat)
+
+- macro: rocketmq_check_properties_file_write_conf
+  condition: (fd.nameraw contains conf and fd.nameraw endswith '.properties')
+
+- rule: Apache Rocketmq CVE-2023-33246 Exploit Attempt
+  desc: >
+    Look for the file ends with '.properties' created under folder 'CONF' upon exploit attempt
+  condition: >
+    rocketmq_check_env_variable and check_file_openat and rocketmq_check_properties_file_write_conf
+  output: CVE-2023-33246 exploit attempt detected (event=%evt.type filepath=%fd.name server_ip=%fd.sip server_port=%fd.sport proto=%fd.l4proto parent=%proc.pname process=%proc.name command=%proc.cmdline container_id=%container.id)
+  priority: WARNING
+  tags: [host, CVE_2023_33246, rocketmq, apache,Mitre_Initial_Access, Mitre_Execution,TA0001, TA0002, T1133, T1059]
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -3425,17 +3425,21 @@
 - macro: rocketmq_check_netty_server
   condition: (proc.name contains NettyServer and proc.env contains ROCKETMQ)
 
+- macro: read_socket
+  condition: evt.type=read and (fd.typechar='4' or fd.typechar='6')
+
 - macro: rocketmq_check_payload_for_conf_update
-  condition: ((fd.typechar='4' or fd.typechar='6') and evt.type=read and evt.buffer contains 'code":25' and evt.buffer contains serializeTypeCurrentRPC and evt.buffer contains JSON)
+  condition: (read_socket and evt.buffer contains 'code":25' and evt.buffer contains serialize)
 
 - rule: Apache Rocketmq CVE-2023-33246 Exploit Detection
   desc: >
     CVE-2023-33246 is exploited at the function used to update configuration & can be detected by looking at the buffer for broker code-25 and serializeTypeCurrentRPC function.
   condition: >
-    rocketmq_check_netty_server and rocketmq_check_payload_for_conf_update and evt.dir=<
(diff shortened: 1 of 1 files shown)


rules/falco_rules.yaml
```diff
@@ -3420,3 +3420,38 @@
 # Application rules have moved to application_rules.yaml. Please look
 # there if you want to enable them by adding to
 # falco_rules.local.yaml.
+
+
+- macro: rocketmq_check_netty_server
+  condition: (proc.name contains NettyServer and proc.env contains ROCKETMQ)
+
+- macro: rocketmq_check_payload_for_conf_update
+  condition: (evt.type=read and evt.buffer contains 'code":25' and evt.buffer contains serializeTypeCurrentRPC and evt.buffer contains JSON)
```
incertum, line 3429: Have you had the opportunity to test this detection in a production environment? If yes have you noticed changes in resource utilization and event drops?

Context: I/O syscalls especially `read` and `write` are quite heavy and the primary concern is that for most adopters the event drop could drastically increase. For this reason existing read and write macros have been disabled https://github.com/falcosecurity/rules/blob/main/rules/falco_rules.yaml#L22C2-L28

All this being said we need to find better ways to support these types of detections, it's a current tooling limitation we are painfully aware.

We could consider disabling this detection by default. CC @falcosecurity/rules-maintainers

rules/falco_rules.yaml
```diff
@@ -3420,3 +3420,38 @@
...
+
+- rule: Apache Rocketmq CVE-2023-33246 Exploit Detection
+  desc: >
+    CVE-2023-33246 is exploited at the function used to update configuration & can be detected by looking at the buffer for broker code-25 and serializeTypeCurrentRPC function.
+  condition: >
+    rocketmq_check_netty_server  and rocketmq_check_payload_for_conf_update
+  output: exploit detected CVE-2023-33246 (event=%evt.type server_ip=%fd.sip server_port=%fd.sport proto=%fd.l4proto parent=%proc.pname process=%proc.name command=%proc.cmdline container_id=%container.id)
+  priority: CRITICAL
+  tags: [host, CVE_2023_33246, rocketmq, apache,Mitre_Initial_Access, Mitre_Execution,TA0001, TA0002, T1133, T1059]
```
incertum, line 3438: `host, container`

as it would fire for both types of workloads, and we are to blame to not yet have this better documented, it's in progress.

rules/falco_rules.yaml
```diff
@@ -3420,3 +3420,38 @@
...
+
+
+- macro: rocketmq_check_env_variable
+  condition: (proc.env contains ROCKETMQ)
+
+- macro: check_file_openat
+  condition: (fd.type=file and evt.type=openat)
```
incertum, line 3445: Could we use an existing macro? https://github.com/falcosecurity/rules/blob/main/rules/falco_rules.yaml#L36C1-L43

incertum (commented):
Thank you @Loginsoft-Research! Left some initial comments.

rules/falco_rules.yaml
hkonduri: > Have you had the opportunity to test this detection in a production environment? If yes have you noticed changes in resource utilization and event drops?q

I have not tested it in a production env. I was looking for a helping hand from the community to make it tested & validated in real prod or beta environments. 

> Context: I/O syscalls especially `read` and `write` are quite heavy and the primary concern is that for most adopters the event drop could drastically increase. For this reason existing read and write macros have been disabled https://github.com/falcosecurity/rules/blob/main/rules/falco_rules.yaml#L22C2-L28

I agree. Unfortunately, the payload was captured only in `read` syscall. Initially, I tried to develop a detection checking the `evt.buffer` without using '-A' switch but however we could not determine the pattern of successful exploitation. 

Would you suggest any workaround where else we can capture the payload? The exploit happens through socket but we did not see much in any sock-based syscalls. 

Another thought is to capture relevant syscalls and compare patterns in the fixed version vs the vulnerable version. This approach would kill a lot of time and we cannot scale it for a large number of detections. 

> All this being said we need to find better ways to support these types of detections, it's a current tooling limitation we are painfully aware.
> 
> We could consider disabling this detection by default. CC @falcosecurity/rules-maintainers

It makes sense to disable it by default and also using a disclaimer in a comment would caution users.

rules/falco_rules.yaml
hkonduri: Sure, somehow I missed it. Will tag container

rules/falco_rules.yaml
hkonduri: Thanks for pointing this out, I shall reuse existing macros. And it is working!

rules/falco_rules.yaml
incertum: > Would you suggest any workaround where else we can capture the payload?

Let's see when I have time for a hacking session to explore, but it seems you already explored various avenues thoroughly. Likely it is something we need to put on the longer-term roadmap to see what else could be done. For Falco 0.36 symlink resolution is prioritized for the next kernel driver revamp.

Would you want to open an issue either here or in the Falco repo for tracking? Would appreciate it!

---


Mind squashing the initial commit and re-pushing with a signoff 🙃 thank you!

rules/falco_rules.yaml
```diff
@@ -3420,3 +3420,35 @@
...
+  condition: (proc.env contains ROCKETMQ)
+
+- macro: rocketmq_check_properties_file_write_conf
+  condition: (fd.nameraw contains conf and fd.nameraw endswith '.properties')
+
+- rule: Apache Rocketmq CVE-2023-33246 Exploit Attempt
+  desc: >
+    Look for the file ends with '.properties' created under folder 'CONF' upon exploit attempt
+  condition: >
+    rocketmq_check_env_variable and open_write and rocketmq_check_properties_file_write_conf and evt.dir=<
```
incertum, line 3451: Once we merge the rules adoption and maturity framework proposal I will work on creating an official style guide and more docs which will include expected review timelines and testing procedures and more in order to set clearer expectations as we all are spread across multiple repos ... just FYI, also I still need to become official approver here first 🙃 

```
open_write and evt.dir=< and rocketmq_check_env_variable and rocketmq_check_properties_file_write_conf 
```

rules/falco_rules.yaml
```diff
@@ -3420,3 +3420,35 @@
 # Application rules have moved to application_rules.yaml. Please look
 # there if you want to enable them by adding to
 # falco_rules.local.yaml.
+
+
+- macro: rocketmq_check_netty_server
+  condition: (proc.name contains NettyServer and proc.env contains ROCKETMQ)
+
+- macro: rocketmq_check_payload_for_conf_update
+  condition: ((fd.typechar='4' or fd.typechar='6') and evt.type=read and evt.buffer contains 'code":25' and evt.buffer contains serializeTypeCurrentRPC and evt.buffer contains JSON)
```
incertum, line 3429: Maybe even worth a new macro `read_socket` or similar for, what is evt.dir here? 

```
evt.type=read and (fd.typechar='4' or fd.typechar='6') ...
```

rules/falco_rules.yaml
```diff
@@ -3420,3 +3420,35 @@
...
+
+- rule: Apache Rocketmq CVE-2023-33246 Exploit Detection
+  desc: >
+    CVE-2023-33246 is exploited at the function used to update configuration & can be detected by looking at the buffer for broker code-25 and serializeTypeCurrentRPC function.
+  condition: >
+    rocketmq_check_netty_server and rocketmq_check_payload_for_conf_update and evt.dir=<
+  output: exploit detected CVE-2023-33246 (event=%evt.type server_ip=%fd.sip server_port=%fd.sport proto=%fd.l4proto parent=%proc.pname process=%proc.name command=%proc.cmdline container_id=%container.id)
```
incertum, line 3436:
```
enabled: false
```

Plus can the entire rule and the respective macros be later commented out similar to how we do it for existing rules that use read? We can do this as last step to first have the CI check syntax ...

incertum:
Related to my comment https://github.com/falcosecurity/rules/pull/85#discussion_r1230144247 could I AB test with you to provide me with a test script to simulate the attack in a VM or other instructions so that I can do quick verifications given reviewers are time constrained?

It's something we could debate and possibly add as future contribution requirement, WDYT?

rules/falco_rules.yaml
Loginsoft-Research: Sure, a style guide definitely helps in the matter of ease of use and readability. In fact, naming conventions when we develop detections with various products are needed in my opinion. Here I used "product_action" as a convention but I am happy to take suggestions from the community. 

Also, I would suggest having a field or tag to know whether the rule detects an anomaly or active exploit/intrusion. I would prefer using tags `anomaly`, `dfir`, `exploit`. 

`Anomaly` - indicates there's an anomaly. (Default falco_rules triggers many anomalies and upon investigation, it results out as false positives)

`DFIR` - indicates there's actually an incident that occurred (which are certain based on intelligence or knowledge base), a few examples cryptominers, ransomware, webshells etc..

`Exploit` - Indicates there's an exploit activity observed (mostly specific to CVE/known exploits)

Given these tags helps me to run only the ones which I needed most based on my exercise. Again, we can work on the appropriate naming conventions. This is just a suggestion coming from experience leading Detection and Engineering team.

hkonduri: Sure, a style guide definitely helps in the matter of ease of use and readability. In fact, naming conventions when we develop detections with various products are needed in my opinion. Here I used "product_action" as a convention but I am happy to take suggestions from the community. 

Also, I would suggest having a field or tag to know whether the rule detects an anomaly or active exploit/intrusion. I would prefer using tags `anomaly`, `dfir`, `exploit`. 

`Anomaly` - indicates there's an anomaly. (Default falco_rules triggers many anomalies and upon investigation, it results out as false positives)

`DFIR` - indicates there's actually an incident that occurred (which are certain based on intelligence or knowledge base), a few examples cryptominers, ransomware, webshells etc..

`Exploit` - Indicates there's an exploit activity observed (mostly specific to CVE/known exploits)

Given these tags helps me to run only the ones which I needed most based on my exercise. Again, we can work on the appropriate naming conventions. This is just a suggestion coming from experience leading Detection and Engineering team.

rules/falco_rules.yaml
Loginsoft-Research: evt.dir is '<' exit events

hkonduri: evt.dir is '<' exit events

hkonduri:
> Related to my comment [#85 (comment)](https://github.com/falcosecurity/rules/pull/85#discussion_r1230144247) could I AB test with you to provide me with a test script to simulate the attack in a VM or other instructions so that I can do quick verifications given reviewers are time constrained?
> 
> It's something we could debate and possibly add as future contribution requirement, WDYT?

Can you use this [container](https://github.com/vulhub/vulhub/tree/master/rocketmq/CVE-2023-33246) to quickly deploy? will this work for you? 
And we are following POC to reproduce the exploit 

- https://github.com/Malayke/CVE-2023-33246_RocketMQ_RCE_EXPLOIT/blob/main/check.py
- https://github.com/hheeyywweellccoommee/CVE-2023-33246-rnkku
- [rocketmq_exploit.zip](https://github.com/falcosecurity/rules/files/11753500/rocketmq_exploit.zip)

I also shared the scap here, you could use sysdig for analysis and falco to test :)

rules/falco_rules.yaml
```diff
@@ -3425,17 +3425,21 @@
 - macro: rocketmq_check_netty_server
   condition: (proc.name contains NettyServer and proc.env contains ROCKETMQ)
 
+- macro: read_socket
+  condition: evt.type=read and (fd.typechar='4' or fd.typechar='6')
+
 - macro: rocketmq_check_payload_for_conf_update
-  condition: ((fd.typechar='4' or fd.typechar='6') and evt.type=read and evt.buffer contains 'code":25' and evt.buffer contains serializeTypeCurrentRPC and evt.buffer contains JSON)
+  condition: (read_socket and evt.buffer contains 'code":25' and evt.buffer contains serialize)
```
hkonduri, line 3432: We had to cut short other strings because 0.35 is not giving desired output as compared to 0.34. However, this condition should still detect the exploit

rules/falco_rules.yaml
```diff
@@ -3425,17 +3425,21 @@
...
 
 - rule: Apache Rocketmq CVE-2023-33246 Exploit Detection
   desc: >
     CVE-2023-33246 is exploited at the function used to update configuration & can be detected by looking at the buffer for broker code-25 and serializeTypeCurrentRPC function.
   condition: >
-    rocketmq_check_netty_server and rocketmq_check_payload_for_conf_update and evt.dir=<
-  output: exploit detected CVE-2023-33246 (event=%evt.type server_ip=%fd.sip server_port=%fd.sport proto=%fd.l4proto parent=%proc.pname process=%proc.name command=%proc.cmdline container_id=%container.id)
+    evt.dir=< and rocketmq_check_netty_server and rocketmq_check_payload_for_conf_update
+#  enabled: false
```
hkonduri, line 3439: We can uncomment it after the review and validation is done

rules/falco_rules.yaml
incertum: Thank you Hari! I'll try to take a look tomorrow, huh oh ok this does not sounds great that things changed between Falco versions ... this is something separate we should look into in the libs repo ...

incertum:
Few more ideas:

In the proposal for the "rules adoption, management, and maturity framework" we are currently finalizing the details. These new processes will include official rule contribution guidelines, rules criteria, as well as style guides.

I wanted to ask if it would be acceptable to wait until these guidelines are published in the upcoming weeks. This way, we will have a formal process in place for tagging and maintaining application-specific rules. 

---


Meanwhile in response to https://github.com/falcosecurity/rules/pull/85#discussion_r1229081268, I have noticed that there might be a general gap in terms of generic behavioral-based detections for RCE originating from Java apps. I believe that introducing such a rule could provide significant value to the community and be of interest to end users who prioritize behavior-based detections.

Since Falco 0.35, we have exposed `proc.aexepath`, which was not available before. This enhancement enables us to perform more robust searches for Java in the parent process lineage, considering that Java applications often have a custom name for `proc.name`.

I am curious to hear your thoughts?

```
spawned_process and proc.name in (shell_binaries) and proc.aexepath endswith java and not ... [some TEMPLATE macros to exclude known conditions that can be customized by adopters]
```

leogr:
/remove-lifecycle rotten

I guess this is still valid. cc @falcosecurity/rules-maintainers

PR #90, opened 2023-06-21
Add filter on event result field in `spawned_process` macro

Biagio-Dipalma:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/area rules

> /area registry

> /area build

> /area documentation

This is a possible improvement of the `spawned_process` macro: basically I'm adding a filter to consider only the successful spawns, not all of them. This will make the detections more precise and avoid useless events. 

Fixes #

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -82,6 +82,9 @@
 - macro: spawned_process
   condition: (evt.type in (execve, execveat) and evt.dir=<)
 
+- macro: successful_spawned_process
+  condition: (evt.type in (execve, execveat) and evt.dir=< and evt.arg.res=0)  
+
 - macro: create_symlink
   condition: (evt.type in (symlink, symlinkat) and evt.dir=<)
 
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -83,7 +83,7 @@
   condition: (evt.type in (execve, execveat) and evt.dir=<)
 
 - macro: successful_spawned_process
-  condition: (evt.type in (execve, execveat) and evt.dir=< and evt.arg.res=0)  
+  condition: (evt.type in (execve, execveat) and evt.dir=< and evt.rawarg.res=0)  
 
 - macro: create_symlink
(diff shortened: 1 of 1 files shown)


Kaizhe:
This is going to impact all the rules using `spawn_process` macro which may cause unexpected results. E.g. some may expect to see the spawn process with errors. I would recommend to create a new macro :)

loresuso:
Agreed with @Kaizhe, this can be really impactful but I do agree that most of the time we are interested in successful `execve`s. I'd create a new macro too, since we still don't have a way to test each rule and we don't know how these changes will affect them. We can start writing new rules with the new macro, and as soon as we introduce some testing, gradually switch to the new one :)

Biagio-Dipalma:
@loresuso / @Kaizhe you're right, I've just added a new macro. TY! :)

loresuso:
Perfect, can you sign-off your commit? So that we can let this in :)

rules/falco_rules.yaml
```diff
@@ -82,6 +82,9 @@
 - macro: spawned_process
   condition: (evt.type in (execve, execveat) and evt.dir=<)
 
+- macro: successful_spawned_process
+  condition: (evt.type in (execve, execveat) and evt.dir=< and evt.arg.res=0)
```
Kaizhe, line 86: I think we should use `evt.rawres` or `evt.rawarg.res` because the `evt.arg` is going to resolute the return value, e.g. `ENOENT`.

Biagio-Dipalma: From my tests the `evt.arg.res` field is not resolved if it's equal to `0`, anyway I'll use the suggested field to be sure that this change will survive to future mappings on this field. Ty again for the precious suggestions

incertum:
@Kaizhe and @loresuso do we have a status update? Are we ok with having unused macros? Perhaps we could comment it, so it serves as example and inspiration? We have done it that way on other places.

leogr:
I believe this needs to be rebased to make the more recent CI jobs run on it.
/assing @LucaGuerra

PR #98, opened 2023-07-12
new(rules): add rule for detecting fileless execution via memfd_create

loresuso:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

/area rules

> /area registry

> /area build

> /area documentation

This PR takes advantage of a new incoming Falco feature for detecting fileless execution via `memfd_create`. This is a very well known defense evasion technique that threat actors might use to execute malware on victim machines. It has been used in the past, but also pretty recently as reported by these two articles:
- https://cybersecurity.att.com/blogs/labs-research/malware-using-new-ezuri-memory-loader
- https://www.wiz.io/blog/pyloose-first-python-based-fileless-attack-on-cloud-workloads
Putting this on wip until the feature lands in Falco. 

Fixes #

### DIFF
--- a/rules/falco-incubating_rules.yaml
+++ b/rules/falco-incubating_rules.yaml
@@ -24,7 +24,7 @@
 
 # Starting with version 8, the Falco engine supports exceptions.
 # However the Falco rules file does not use them by default.
-- required_engine_version: 17
+- required_engine_version: 26
 
 - macro: open_write
   condition: (evt.type in (open,openat,openat2) and evt.is_open_write=true and fd.typechar='f' and fd.num>=0)
--- a/rules/falco-incubating_rules.yaml
+++ b/rules/falco-incubating_rules.yaml
@@ -1221,6 +1221,7 @@
   output: Exfiltrating Artifacts via Kubernetes Control Plane (file=%fd.name evt_type=%evt.type user=%user.name user_uid=%user.uid user_loginuid=%user.loginuid process=%proc.name proc_exepath=%proc.exepath parent=%proc.pname command=%proc.cmdline terminal=%proc.tty exe_flags=%evt.arg.flags %container.info)
   priority: NOTICE
   tags: [maturity_incubating, container, filesystem, mitre_exfiltration, TA0010]
+
 - list: known_memfd_execution_binaries
   items: []
 
@@ -1229,17 +1230,14 @@
 
 - rule: Fileless execution via memfd_create
   desc: >
-    Detect if a binary is executed from memory using the memfd_create technique.
-    This is a well-known defense evasion technique for executing malware on victim machine without storing payload on disk and to avoid leaving traces about what has been executed. 
-    Adopters can whitelist processes that may use fileless execution for benign purposes by adding itmes to the list `known_memfd_execution_processes`.
-    It is worth underline that this technique is used by runc to fix CVE-2019-5736. For this reason, the macro `known_memfd_execution_processes` take this into account.
+    Detect if a binary is executed from memory using the memfd_create technique. This is a well-known defense evasion 
+    technique for executing malware on a victim machine without storing the payload on disk and to avoid leaving traces 
+    about what has been executed. Adopters can whitelist processes that may use fileless execution for benign purposes 
+    by adding items to the list known_memfd_execution_processes.
   condition: >
     spawned_process
     and proc.is_exe_from_memfd=true
     and not known_memfd_execution_processes
-  output: > 
-    Detected fileless execution via memfd_create (user=%user.name user_loginuid=%user.loginuid user_uid=%user.uid comm=%proc.cmdline exe=%proc.exe container_id=%container.id
-    image=%container.image.repository proc.name=%proc.name proc.sname=%proc.sname proc.pname=%proc.pname proc.aname[2]=%proc.aname[2] exe_flags=%evt.arg.flags
-    proc.exepath=%proc.exepath proc.cwd=%proc.cwd proc.tty=%proc.tty container.start_ts=%container.start_ts proc.sid=%proc.sid proc.vpgid=%proc.vpgid evt.res=%evt.res)
+  output: Fileless execution via memfd_create (container_start_ts=%container.start_ts proc_cwd=%proc.cwd evt_res=%evt.res proc_sname=%proc.sname gparent=%proc.aname[2] evt_type=%evt.type user=%user.name user_uid=%user.uid user_loginuid=%user.loginuid process=%proc.name proc_exepath=%proc.exepath parent=%proc.pname command=%proc.cmdline terminal=%proc.tty exe_flags=%evt.arg.flags %container.info)
   priority: CRITICAL
   tags: [maturity_incubating, host, container, process, mitre_defense_evasion, T1620]
--- a/rules/falco-incubating_rules.yaml
+++ b/rules/falco-incubating_rules.yaml
@@ -1221,3 +1221,25 @@
   output: Exfiltrating Artifacts via Kubernetes Control Plane (file=%fd.name evt_type=%evt.type user=%user.name user_uid=%user.uid user_loginuid=%user.loginuid process=%proc.name proc_exepath=%proc.exepath parent=%proc.pname command=%proc.cmdline terminal=%proc.tty exe_flags=%evt.arg.flags %container.info)
   priority: NOTICE
   tags: [maturity_incubating, container, filesystem, mitre_exfiltration, TA0010]
+- list: known_memfd_execution_binaries
+  items: []
+
+- macro: known_memfd_execution_processes
+  condition: (proc.name in (known_memfd_execution_binaries))
+
+- rule: Fileless execution via memfd_create
+  desc: >
+    Detect if a binary is executed from memory using the memfd_create technique.
+    This is a well-known defense evasion technique for executing malware on victim machine without storing payload on disk and to avoid leaving traces about what has been executed. 
+    Adopters can whitelist processes that may use fileless execution for benign purposes by adding itmes to the list `known_memfd_execution_processes`.
+    It is worth underline that this technique is used by runc to fix CVE-2019-5736. For this reason, the macro `known_memfd_execution_processes` take this into account.
+  condition: >
+    spawned_process
+    and proc.is_exe_from_memfd=true
+    and not known_memfd_execution_processes
+  output: > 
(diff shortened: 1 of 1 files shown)


rules/falco_rules.yaml
```diff
@@ -3416,6 +3416,21 @@
...
+  condition: >
+    spawned_process
+    and proc.is_exe_from_memfd=true
+  output: > 
+    Detected fileless execution via memfd_create (user=%user.name user_loginuid=%user.loginuid user_uid=%user.uid comm=%proc.cmdline exe=%proc.exe container_id=%container.id
+    image=%container.image.repository proc.name=%proc.name proc.sname=%proc.sname proc.pname=%proc.pname proc.aname[2]=%proc.aname[2] exe_flags=%evt.arg.flags
+    proc.exe_ino=%proc.exe_ino proc.exe_ino.ctime=%proc.exe_ino.ctime proc.exe_ino.mtime=%proc.exe_ino.mtime proc.exe_ino.ctime_duration_proc_start=%proc.exe_ino.ctime_duration_proc_start
+    proc.exepath=%proc.exepath proc.cwd=%proc.cwd proc.tty=%proc.tty container.start_ts=%container.start_ts proc.sid=%proc.sid proc.vpgid=%proc.vpgid evt.res=%evt.res)
+  priority: CRITICAL
+  tags: [mitre_defense_evasion, T1620]
```
incertum, line 3432:
```suggestion
  tags: [maturity_incubating, host, container, mitre_defense_evasion, T1620]
```

rules/falco_rules.yaml
```diff
@@ -3416,6 +3416,21 @@
...
+  desc: 
+    Detect if a binary is executed from memory using the memfd_create technique.
+    This is a well-known defense evasion technique for executing malware on victim machine without storing payload on disk.
+  condition: >
+    spawned_process
+    and proc.is_exe_from_memfd=true
+  output: > 
+    Detected fileless execution via memfd_create (user=%user.name user_loginuid=%user.loginuid user_uid=%user.uid comm=%proc.cmdline exe=%proc.exe container_id=%container.id
+    image=%container.image.repository proc.name=%proc.name proc.sname=%proc.sname proc.pname=%proc.pname proc.aname[2]=%proc.aname[2] exe_flags=%evt.arg.flags
+    proc.exe_ino=%proc.exe_ino proc.exe_ino.ctime=%proc.exe_ino.ctime proc.exe_ino.mtime=%proc.exe_ino.mtime proc.exe_ino.ctime_duration_proc_start=%proc.exe_ino.ctime_duration_proc_start
```
incertum, line 3429: All `proc.exe_ino.*` should be null in this scenario, just remove.

rules/falco_rules.yaml
```diff
@@ -3416,6 +3416,21 @@
   priority: CRITICAL 
   tags: [container, mitre_persistence, TA0003]
 
+- rule: Fileless execution via memfd_create
+  desc: 
+    Detect if a binary is executed from memory using the memfd_create technique.
+    This is a well-known defense evasion technique for executing malware on victim machine without storing payload on disk.
+  condition: >
+    spawned_process
+    and proc.is_exe_from_memfd=true
```
incertum, line 3425: Any tuning ideas, this likely can be a bit noisy in production?

incertum:
@loresuso we can soon start with merging rules that make the rules file incompatible with the current Falco as the next release is planned for end of Sep.

incertum:
After merging this rule at `maturity_incubating` we can bump to `maturity_stable` in a follow up PR. That way we honor our new guidelines. CC @darryk10 @leogr

leogr:
> After merging this rule at `maturity_incubating` we can bump to `maturity_stable` in a follow up PR. That way we honor our new guidelines. CC @darryk10 @leogr

:+1: 

Rules files have been split. @loresuso could you rebase this PR and address the above suggestions, please? :pray:

rules/falco_rules.yaml
loresuso: Added some tuning with the classic whitelisting methods. Also, added a description about CVE-2019-5736, which was solved by runc using memfd_create

rules/falco-incubating_rules.yaml
```diff
@@ -1205,4 +1205,27 @@
...
+  desc: >
+    Detect if a binary is executed from memory using the memfd_create technique.
+    This is a well-known defense evasion technique for executing malware on victim machine without storing payload on disk and to avoid leaving traces about what has been executed. 
+    Adopters can whitelist processes that may use fileless execution for benign purposes by adding itmes to the list `known_memfd_execution_processes`.
+    It is worth underline that this technique is used by runc to fix CVE-2019-5736. For this reason, the macro `known_memfd_execution_processes` take this into account.
+  condition: >
+    spawned_process
+    and proc.is_exe_from_memfd=true
+    and not known_memfd_execution_processes
+  output: >
```
incertum, line 1226:
```suggestion
  output: Fileless execution via memfd_create (container_start_ts=%container.start_ts proc_cwd=%proc.cwd evt_res=%evt.res proc_sname=%proc.sname gparent=%proc.aname[2] evt_type=%evt.type user=%user.name user_uid=%user.uid user_loginuid=%user.loginuid process=%proc.name proc_exepath=%proc.exepath parent=%proc.pname command=%proc.cmdline terminal=%proc.tty exe_flags=%evt.arg.flags %container.info)
```

incertum: I preserved the eligible unique fields you proposed except for the ID related ones, see Style Guide https://falco.org/docs/rules/style-guide/#output-fields.

rules/falco-incubating_rules.yaml
```diff
@@ -1205,4 +1205,27 @@
...
+  tags: [maturity_incubating, container, filesystem, process, mitre_discovery, T1083]
+
+- list: known_memfd_execution_binaries
+  items: []
+
+- macro: known_memfd_execution_processes
+  condition: (container_entrypoint or (proc.name in known_memfd_execution_binaries))
+
+- rule: Fileless execution via memfd_create
+  desc: >
```
incertum, line 1217:
```suggestion
  desc: >
    Detect if a binary is executed from memory using the memfd_create technique. This is a well-known defense evasion 
    technique for executing malware on a victim machine without storing the payload on disk and to avoid leaving traces 
    about what has been executed. Adopters can whitelist processes that may use fileless execution for benign purposes 
    by adding items to the list known_memfd_execution_processes. For example, runc uses this technique to address 
    CVE-2019-5736. Therefore, we have already included runc in the tuning macro known_memfd_execution_processes.
```

rules/falco-incubating_rules.yaml
```diff
@@ -1205,4 +1205,27 @@
     and not proc.name in (known_binaries_to_read_environment_variables_from_proc_files)
   output: Environment variables were retrieved from /proc files (file=%fd.name gparent=%proc.aname[2] ggparent=%proc.aname[3] gggparent=%proc.aname[4] evt_type=%evt.type user=%user.name user_uid=%user.uid user_loginuid=%user.loginuid process=%proc.name proc_exepath=%proc.exepath parent=%proc.pname command=%proc.cmdline terminal=%proc.tty exe_flags=%evt.arg.flags %container.info)
   priority: WARNING
-  tags: [maturity_incubating, container, filesystem, process, mitre_discovery, T1083]
\ No newline at end of file
+  tags: [maturity_incubating, container, filesystem, process, mitre_discovery, T1083]
+
+- list: known_memfd_execution_binaries
+  items: []
```
incertum, line 1211:
```suggestion
  items: [runc]
```

???

rules/falco-incubating_rules.yaml
incertum: Some super minor language edits and attempting to make the last sentences clearer, WDYT?

incertum (commented):
@loresuso few minor change requests.

In addition, adding this to the tracking issue https://github.com/falcosecurity/rules/issues/138 as it counts towards modernizing Falco rules.

incertum (commented):
/hold

We need to bump the required engine version. @leogr @LucaGuerra do we bump the rules engine uniformly in all rules files? I would vote for that!

loresuso:
IMO it makes sense to uniformly bump the engine version. In this way, when we graduate rules, we don't need to bother or remember how the version should be bumped, and just move the rule between files, but I'm open to discuss!

leogr:
> /hold
> 
> We need to bump the required engine version. @leogr @LucaGuerra do we bump the rules engine uniformly in all rules files? I would vote for that!

What's the benefit of having the same `required_engine_version`? 
The only two consequences I see
- We will have to bump the minor version in all rules files https://github.com/falcosecurity/rules/blob/main/RELEASE.md#versioning-a-ruleset
- We will make all files not compatible with a previous Falco engine (even if they are)

Moreover, the CI should already be able to check (and inform us) when an engine version bump is required. Indeed, this jobs matrix indicates that `falco-incubating_rules.yaml` is no longer compatible with `0.35.0` because the missing field
Screenshot from 2023-09-14 11-38-47


In the end, I don't have a strong opinion and am open to any solution. I just don't see the benefit of doing this (perhaps am I missing something? :thinking: )

In case we decide to bump all files, I still would avoid bumping [falco-deprecated_rules.yaml](https://github.com/falcosecurity/rules/blob/main/rules/falco-deprecated_rules.yaml), since it does not make much sense to me.

incertum (commented):
Looks great now, will approve once we have decided on how to bump the required engine version @leogr, thanks!

incertum:
@leogr will the required engine version be overriden in the order the rules files are loaded? If we don't bump in all files, we at least should make sure the highest required version in any files has the final say?

leogr:
> @leogr will the required engine version be overriden in the order the rules files are loaded? If we don't bump in all files, we at least should make sure the highest required version in any files has the final say?

The `required_engine_version` applies at the single file level only (@jasondellaluce correct me if I'm wrong). It acts as a validation condition when loading the rules files, so it won't be overridden by other rules files. Basically, files will be validated one by one, the first failure will stop the whole process, regardless of the loading order.

incertum:
> > @leogr will the required engine version be overriden in the order the rules files are loaded? If we don't bump in all files, we at least should make sure the highest required version in any files has the final say?
> 
> The `required_engine_version` applies at the single file level only (@jasondellaluce correct me if I'm wrong). It acts as a validation condition when loading the rules files, so it won't be overridden by other rules files. Basically, files will be validated one by one, the first failure will stop the whole process, regardless of the loading order.

ok then let's just bump it in the respective file for now, thanks!

incertum (approved):
/approve

Thanks a bunch @loresuso 🚀!

PR #162, opened 2023-09-11
fix(build): update latest even if higher prereleases are present

LucaGuerra:
**What type of PR is this?**

/kind bug

/area build

If rules marked as 1.0.0, 1.0.1 and 2.0.0-rc1 are present, when 1.0.2 is released it should still be latest because 2.0.0 is not out yet. Fix the build process to consider this case (note the attached test)

Fixes #


incertum (approved):
/approve

PR #180, opened 2023-09-29
update: fix deprecated rules and refactor rules checker to support newest falco output

jasondellaluce:
**What type of PR is this?**

kind feature

/kind bug

/kind cleanup

/area rules

/area build

/area maturity-deprecated

This bumps the dependency of the rules checker tool to falcosecurity/testing to support its newest definitions, which are compatible with the latest changes of the Falco `-L` output.

Also, this fixes the deprecated ruleset by rolling back its required engine version scheme to non-semver. Semantic versioning is supported only after engine version 0.26.0. Using a semantic version on older version will be incompatible with older Falcos, thus defeating the purposes of having < 26 version requirement.


FedeDP (approved):
/approve

Issue #214, opened 2024-01-05
Wrong usage of `evt.arg.*` / `evt.rawarg.*` when more than one event is involved


Andreagit97:
In some rules like `Set Setuid or Setgid bit` we use some filter checks like `fd=%evt.arg.fd`. These filter checks are event-specific so using them with a condition like `evt.type in (chmod, fchmod, fchmodat)` means that `chmod` and `fchmodat` don't have it and will always return an `<NA>` while  `fchmod` has it.

I'm not sure this is what we want, returning `<NA>` by default doesn't seem the right choice. IMO we should limit the usage of `evt.arg.*` / `evt.rawarg.*` when only one event is involved in the condition and a precise direction is set (e.g. `evt.type=open and evt.dir=<`). The direction is necessary because enter and exit events have different parameters!

incertum:
Cross-linked the issue to the feedback tracking https://github.com/falcosecurity/rules/issues/176

PR #215, opened 2024-01-05
cleanup: remove `evt.arg.*` fields when always return `<NA>`

Andreagit97:
**What type of PR is this?**

/kind cleanup

/area rules

This PR removes all `evt.arg.*` fields that will be always resolved as `<NA>`. 

Related to this: https://github.com/falcosecurity/rules/issues/214


rules/falco-deprecated_rules.yaml
```diff
@@ -87,7 +87,7 @@
     and ssh_port 
     and not allowed_ssh_hosts
   enabled: false
-  output: Disallowed SSH Connection (connection=%fd.name lport=%fd.lport rport=%fd.rport fd_type=%fd.type fd_proto=fd.l4proto evt_type=%evt.type user=%user.name user_uid=%user.uid user_loginuid=%user.loginuid process=%proc.name proc_exepath=%proc.exepath parent=%proc.pname command=%proc.cmdline terminal=%proc.tty exe_flags=%evt.arg.flags %container.info)
+  output: Disallowed SSH Connection (connection=%fd.name lport=%fd.lport rport=%fd.rport fd_type=%fd.type fd_proto=fd.l4proto evt_type=%evt.type user=%user.name user_uid=%user.uid user_loginuid=%user.loginuid process=%proc.name proc_exepath=%proc.exepath parent=%proc.pname command=%proc.cmdline terminal=%proc.tty %container.info)
```
Andreagit97, line 90: events like `accept,accept4,listen,connect` don't have a `flag` param so `exe_flags=%evt.arg.flags` will be always evaluated to `<NA>`. Same thing in many of the other cases in this PR

rules/falco-incubating_rules.yaml
```diff
@@ -1040,6 +1044,11 @@
 - macro: user_known_set_setuid_or_setgid_bit_conditions
   condition: (never_true)
 
+# todo!: the usage of `evt.arg*` filter check in the output should be avoided
```
Andreagit97, line 1047: See https://github.com/falcosecurity/rules/issues/214

rules/falco_rules.yaml
```diff
@@ -984,8 +984,8 @@
     rule "Read sensitive file untrusted".
   condition: >
     create_hardlink 
-    and (evt.arg.oldpath in (sensitive_file_names))
-  output: Hardlinks created over sensitive files (target=%evt.arg.target linkpath=%evt.arg.linkpath evt_type=%evt.type user=%user.name user_uid=%user.uid user_loginuid=%user.loginuid process=%proc.name proc_exepath=%proc.exepath parent=%proc.pname command=%proc.cmdline terminal=%proc.tty exe_flags=%evt.arg.flags %container.info)
+    and (fs.path.source in (sensitive_file_names))
+  output: Hardlinks created over sensitive files (target=%fs.path.target linkpath=%fs.path.source evt_type=%evt.type user=%user.name user_uid=%user.uid user_loginuid=%user.loginuid process=%proc.name proc_exepath=%proc.exepath parent=%proc.pname command=%proc.cmdline terminal=%proc.tty %container.info)
```
Andreagit97, line 988: `evt.arg.target` and `evt.arg.linkpath` don't exist for `link` and `linkat`

rules/falco_rules.yaml
```diff
@@ -998,11 +998,11 @@
     and privilege escalation (CVE-2020-14386) by an attacker. Noise can be reduced by using the user_known_packet_socket_binaries
     template list.
   condition: > 
-    evt.type=socket 
+    evt.type=socket and evt.dir=>
```
Andreagit97, line 1001: `evt.arg.domain` is only defined for the enter event

rules/falco_rules.yaml
```diff
@@ -998,11 +998,11 @@
...
     and container 
-    and evt.arg[0] contains AF_PACKET 
+    and evt.arg.domain contains AF_PACKET
```
Andreagit97, line 1003: just for clarity

rules/falco_rules.yaml
incertum: CC @darryk10 and @loresuso to also confirm, thanks!

rules/falco_rules.yaml
incertum: great catch

rules/falco_rules.yaml
incertum: Actually this was a refactor mistake (by me since I had to do it all), I would instead suggest to revert to the old correct outputs: `target=%evt.arg.oldpath linkpath=%evt.arg.newpath`, see in the git history https://github.com/falcosecurity/rules/blob/dc90ca5e179fcf11e6a9b8486e4f23dacb613087/rules/falco_rules.yaml#L2808C126-L2808C175

incertum: Because using the new `fs.path*` fields would break things and it would be inconsistent with the symlink rule.

rules/falco-incubating_rules.yaml
Andreagit97: uhm here I would prefer to remove the `flags=%...` output field unless it is not really necessary in the rule. I would reduce the output fields in our rules as much as possible, so if one field is not strictly necessary for the scope of the rule, I prefer to remove it. I have no strong opinions here, this was just my thought

rules/falco_rules.yaml
Andreagit97: done, thank you for the feedback!

rules/falco-incubating_rules.yaml
incertum: ACK, yes I have the website PR already up to make this clear in the Style Guide.

I'll open a follow up PR to rename exe_flags to flags (unrelated to this PR) https://github.com/falcosecurity/rules/pull/215/files#r1443163428

ACK re https://github.com/falcosecurity/rules/pull/215/files#r1444885132

incertum (approved):
/approve

PR #216, opened 2024-01-08
update(deprecated_rules): use SemVer `required_engine_version`

Andreagit97:
**What type of PR is this?**

/kind cleanup

/area rules

/area maturity-deprecated

This PR adapts the deprecated rule file to use a SemVer-compatible `required_engine_version`. This change would avoid a deprecation warning in Falco 0.37.0 (see https://github.com/falcosecurity/falco/pull/2992)

### DIFF
--- a/rules/falco-deprecated_rules.yaml
+++ b/rules/falco-deprecated_rules.yaml
@@ -25,7 +25,7 @@
 
 # Starting with version 8, the Falco engine supports exceptions.
 # However the Falco rules file does not use them by default.
-- required_engine_version: 17
+- required_engine_version: 0.17.0
 
 # This macro `never_true` is used as placeholder for tuning negative logical sub-expressions, for example
 # - macro: allowed_ssh_hosts


FedeDP:
Makes sense to me.
+1

Andreagit97:
/hold 
Not sure we want to do this now

Andreagit97:
we need to understand the backward compatibility issue, if we want to change all our engine versions to a SemVer compatible number I would propose doing it for all our rulesets, also the plugin ones... let's hold this until we understand what we want to do

incertum (approved):
/approve

/unhold

Yes we agreed already in another PR.

leogr:
Are we sure we want this? :thinking:  It will make this rulesfile not usable anymore with previous Falco version.

Andreagit97:
yeah, that was my concern about why it was on hold... btw I can revert it if we decide to keep the non SemVer version

incertum:
> Sorry, I'm late... anyway Are we sure we want this? 🤔 It will make this rulesfile not usable anymore with previous Falco versions.

I think we should cleanly cut over to semVer else it's gonna be a real pain for us maintainers of this repo. In fact I would even suggest we should bump all to the latest version.

I have a readme PR update ready that states this semVer update and that all files are not anymore compatible with previous Falco versions.

It's not ideal, but if we don't do this cutover consistently things won't get better.

leogr:
> > Sorry, I'm late... anyway Are we sure we want this? 🤔 It will make this rulesfile not usable anymore with previous Falco versions.
> 
> I think we should cleanly cut over to semVer else it's gonna be a real pain for us maintainers of this repo. In fact I would even suggest we should bump all to the latest version.
> 
> I have a readme PR update ready that states this semVer update and that all files are not anymore compatible with previous Falco versions.
> 
> It's not ideal, but if we don't do this cutover consistently things won't get better.

If so, we have ti bump the Major version when releasing, at least. Otherwise, we break users using auto update by falcoctl (because switching to semver Is a backward incompatibile change). Make sense?

cc @falcosecurity/rules-maintainers @loresuso @Andreagit97

loresuso:
What about just documenting it?
- Updating everything to 0.26.0 is not correct, because we are not actually requesting new features of the language, we are just doing it's convenient. Agreed that wouldn't cause any harm (that I can see at least). If so yes, the rules have to be a bumped major, as we already agreed to do also for the other rules (this is crucial).
- Requesting 0.17.0 makes no sense, as this implies asking for a newer Falco (older Falcos around can't read the semver value) but the Rules could be totally read by older Falco

So we can either choose option 1 or just draw something like this:

```
-------20--------21---------...---------25----------0.26.0-------->
```

as allowed values for the engine version

PR #208, opened 2023-12-19
Update: bpf cmd format

Rohith-Raju:
**What type of PR is this?**

/kind bug

**Any specific area of the project related to this PR?**

/area rules

**Proposed rule [maturity level](https://github.com/falcosecurity/rules/blob/main/CONTRIBUTING.md#maturity-levels)**

/area maturity-sandbox

**What this PR does / why we need it**: 

This PR updates new bpf cmd format based on changes implemented in https://github.com/falcosecurity/libs/pull/1545

**Which issue(s) this PR fixes**: 

Fixes #203 

**Special notes for your reviewer**:

PR #231, opened 2024-03-01
update(sandbox): Added systemd to list of programs ignored when using BPF.

petterreinholdtsen:
On Fedora 39, events like this is logged:

  Notice BPF Program  Not Profiled (bpf_cmd=5 evt_type=bpf user=root
  user_uid=0 user_loginuid=-1 process=systemd proc_exepath=/usr/lib/systemd/systemd
  parent=<NA> command=systemd --switched-root --system --deserialize=36 rhgb terminal=0
  exe_flags=<NA> container_id=host container_name=host)

This is just normal systemd behaviour, and not really interesting to log.  The provided patch include systemd in the list of processes allowed to use BPF without any event logged.

/kind bug
/area rules
/area maturity-sandbox

### DIFF
--- a/rules/falco-sandbox_rules.yaml
+++ b/rules/falco-sandbox_rules.yaml
@@ -1699,7 +1699,7 @@
   tags: [maturity_sandbox, container, filesystem, mitre_initial_access, T1611]
 
 - list: bpf_profiled_binaries
-  items: [falco, bpftool]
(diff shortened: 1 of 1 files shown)


incertum (approved):
/approve

Thanks, please note that this rule in our main branch just became incompatible with the latest Falco release through those changes: https://github.com/falcosecurity/rules/pull/208

PR #238, opened 2024-03-22
rules yaml files updated as per yaml linting conditions

h4l0gen:
**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind feature

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

/kind failing-test

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

/area rules

> /area registry

> /area build

> /area documentation

**Proposed rule [maturity level](https://github.com/falcosecurity/rules/blob/main/CONTRIBUTING.md#maturity-levels)**

> Uncomment one (or more) `/area <>` lines (only for PRs that add or modify rules):

> /area maturity-stable

> /area maturity-incubating

> /area maturity-sandbox

> /area maturity-deprecated

**What this PR does / why we need it**:
to pass yaml lint workflow test.
**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

PR #244, opened 2024-04-26
fix: the correct usage is `<NA>` not `N/A`

Andreagit97:
**What type of PR is this?**

/kind bug

/area rules

/area maturity-incubating

/area maturity-sandbox

Debugging some issues in Falco CI https://github.com/falcosecurity/falco/actions/runs/8845222090?pr=3177 I faced this inconsistency. The rule `Non sudo setuid` was triggered with `user.name=<NA>`  because the macro `known_user_in_container` checks for `N/A`. This PR fixes the usages of `N/A`


darryk10 (approved):
Thanks @Andreagit97!

Andreagit97:
I will check if we need to keep both `N/A` and `<NA>` or `<NA>` is enough

rules/falco-incubating_rules.yaml
```diff
@@ -769,7 +769,7 @@
 # https://github.com/draios/sysdig/issues/954). So in that case, allow
 # a setuid.
 - macro: known_user_in_container
-  condition: (container and user.name != "N/A")
+  condition: (container and user.name != <NA>)
```
Andreagit97, line 772: possible alternatives
```suggestion
  condition: (container and not user.name in ("<NA>","N/A"))
```

incertum (dismissed):
/approve

Thanks and let's only consistently use the correct version.

Andreagit97:
Uhm `Yamllint Github Actions / Yamllint` suggests that the line length is too long, but this is true for almost all the lines of the files... so probably we should fix it in one shot and not here

leogr (dismissed):

leogr:
> Uhm `Yamllint Github Actions / Yamllint` suggests that the line length is too long, but this is true for almost all the lines of the files... so probably we should fix it in one shot and not here

you can ignore this for now, since here's an ongoing discussion https://github.com/falcosecurity/rules/pull/238

rules/falco-incubating_rules.yaml
```diff
@@ -769,7 +769,7 @@
 # https://github.com/draios/sysdig/issues/954). So in that case, allow
 # a setuid.
 - macro: known_user_in_container
-  condition: (container and not user.name in ("<NA>","N/A"))
+  condition: (container and not user.name in ("<NA>","N/A",""))
```
Andreagit97, line 772: also `""` seems a legit value for `user.name`

leogr (approved):

PR #245, opened 2024-05-13
fix: change CVE-2024-3094 to match liblzma contain instead of endswith

apsega:
**What type of PR is this?**

/kind bug
/area rules
/area maturity-incubating

Some of the docs in the wild of how to reproduce XZ vulnerability relies on patching `liblzma` and having the filename with `patch` suffix, for example - `liblzma.so.5.6.1.patch`, thus this Falco rule doesn't pick this up. Changing rule to match `contains` string resolves the issue.

Some examples of who's using `liblzma.so.5.6.1.patch` name:
- https://github.com/amlweems/xzbot/tree/main/assets
- https://github.com/r0binak/xzk8s/blob/main/Dockerfile#L11

### DIFF
--- a/rules/falco-incubating_rules.yaml
+++ b/rules/falco-incubating_rules.yaml
@@ -1280,7 +1280,7 @@
   condition: > 
     open_read and 
     proc.name=sshd and 
-    (fd.name endswith "liblzma.so.5.6.0" or fd.name endswith "liblzma.so.5.6.1")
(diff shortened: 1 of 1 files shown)


darryk10 (approved):
Thanks, it looks good to me!

leogr (approved):
/milestone falco-0.38-rules

PR #260, opened 2024-08-15
Ignore common UDP ports

jackmtpt:
443 (http3) and 88 (kerberos) are expected to see UDP traffic

**What type of PR is this?**

> /kind feature

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/area rules

> /area registry

> /area build

> /area documentation

> /area maturity-stable

/area maturity-incubating

> /area maturity-sandbox

> /area maturity-deprecated

Fixes #

### DIFF
--- a/rules/falco-incubating_rules.yaml
+++ b/rules/falco-incubating_rules.yaml
@@ -734,7 +734,7 @@
   items: [0, 9, 80, 3306]
 
 - list: expected_udp_ports
-  items: [53, openvpn_udp_ports, l2tp_udp_ports, statsd_ports, ntp_ports, test_connect_ports]
+  items: [53, 443, 88, openvpn_udp_ports, l2tp_udp_ports, statsd_ports, ntp_ports, test_connect_ports]
 
 - macro: expected_udp_traffic
   condition: fd.port in (expected_udp_ports)


incertum:
Thank you 🎉 , we will check on it soon!

jackmtpt:
Looks like the linting job failed for a bunch of files that weren't modified in this PR?

incertum:
> Looks like the linting job failed for a bunch of files that weren't modified in this PR?

Yes this is currently always failing as we have not yet reached a consensus wrt to the best linting approach. This CI check is also not required to pass atm.

Re the changes -- I don't see problems with the changes as in many rules we are a bit more conservative. This rule is an "incubating" rule and highly environment specific wrt what is considered unexpected, so it's fair to expect adopters to adjust this rule for their environment anyways.

@darryk10 could you help with the review? Thanks.

[Some of the reviewers may still be on PTO and this review may be delayed].

leogr:
cc @loresuso

darryk10:
Hi, 
I think both are valid use cases that we might need to whitelist to reduce possible noise.
I wondered if we could whitelist the use case using a port and process using it, instead of removing stopping monitoring the overall port.

jackmtpt:
To be honest we probably would not bother whitelisting (port, app) pairs in these rules; it's too much duplication from network policies/firewall rules.

leogr:
/remove-lifecycle stale

cc @darryk10 @loresuso

jackmtpt:
We don't use Falco anymore so I no longer care about this. Feel free to just close it.

PR #268, opened 2025-01-22
rules: add runc to known_memfd_execution_binaries

ytsssun:
**What type of PR is this?**

> /kind feature

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/area rules

> /area registry

> /area build

> /area documentation

> /area maturity-stable

> /area maturity-incubating

/area maturity-sandbox

> /area maturity-deprecated

This PR is to extend the `known_memfd_execution_binaries` with runc since in 1.1.15 of runc, it introduced a behavior which would use the memfd approach to execute runc binary. This PR is to include the rule for it so that we don't falsely flag the runc binary.

Fixes #3444

I can use come help confirming the change. I was able to verify the override works, see the [proof](https://github.com/falcosecurity/falco/issues/3444#issuecomment-2589906819).

But I would need help on verifying that via the first party rules for falco.

### DIFF
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -1230,10 +1230,15 @@
   tags: [maturity_stable, host, container, network, process, mitre_execution, T1059]
 
 - list: known_memfd_execution_binaries
-  items: []
+  items: [runc]
 
 - macro: known_memfd_execution_processes
-  condition: (proc.name in (known_memfd_execution_binaries))
+  condition: > 
+    (proc.name in (known_memfd_execution_binaries))
+    or (proc.pname in (known_memfd_execution_binaries))
+    or (proc.exepath = "memfd:runc_cloned:/proc/self/exe")
+    or (proc.exe = "memfd:runc_cloned:/proc/self/exe")
+
 
 - rule: Fileless execution via memfd_create
   desc: >


rules/falco_rules.yaml
```diff
@@ -1230,10 +1230,14 @@
   tags: [maturity_stable, host, container, network, process, mitre_execution, T1059]
 
 - list: known_memfd_execution_binaries
-  items: []
+  items: [runc]
 
 - macro: known_memfd_execution_processes
-  condition: (proc.name in (known_memfd_execution_binaries))
+  condition: > 
+    (proc.name in (known_memfd_execution_binaries))
```
loresuso, line 1237: Hi there! thanks for opening the PR. 
I am not sure though this will work since:
- when we execute from memfd, usually we use the proc filesystem reference to the fd to do so. The process name will be a numeric value such as "3"
- IIRC, Falco doesn't resolve exepath and exe that way
Possibly, the best option would be to tune the rule by using ancestors information, such as proc.pname and proc.aname

ytsssun:
Hi @loresuso , if you see my original [repro and fix](https://github.com/falcosecurity/falco/issues/3444#issue-2779282992), this worked for me.
```
customRules:
  rules-allow-runc-memfd.yaml: |-
    - list: known_memfd_execution_binaries
      items: [runc]
      override:
        items: append
```

There indeed was a case where the proc.name is numerical value. And in that case this worked for them.
```
customRules:
  rules-allow-runc-memfd.yaml: |-
    # Add runc to known binaries
    - list: known_memfd_execution_binaries
      items: [runc]
      override:
        items: append

    # Add specific path conditions to existing macro
    - macro: known_memfd_execution_processes
      condition: or (proc.exepath = "memfd:runc_cloned:/proc/self/exe") or (proc.exe = "memfd:runc_cloned:/proc/self/exe")
      override:
        condition: append
```

The implementation covers both cases. Do you think I still need to update it to include "proc.pname and proc.aname"?

ytsssun:
> I would still introduce the exception considering the lineage of the process, so adding proc.pname or proc.aname to match this exact use case accordingly. This will make the rule better, since in theory nobody is preventing an attacker to name his own memfd "runc_cloned", bypassing the rule. If the information about the ancestors is there, the rule turns out to be stronger.

I appended the condition to also check for `proc.pname` in the match so that it is more comprehensive.

> I'll also write this in case we need to remember it, it looks like runc recently dropped the memfd execution in more modern systems in favor of an overlayfs sealed execution for performance reasons [opencontainers/runc#4448](https://github.com/opencontainers/runc/pull/4448)

Looking at the [PR commit](https://github.com/opencontainers/runc/pull/4448/commits/8cfbccb6d99a1bcfad5b16fb3f0bbca97cb5be4c), it looks like this is indeed the optimization they included later on. This change is only included in the `>1.2.0` versions and is not backported to the 1.1.x versions. For now the affected version would be 1.1.15 and any future 1.1.x versions.

loresuso (approved):
LGTM, thanks!

leogr (approved):

jcchavezs:
@leogr can we get this in a release even if beta?

leogr:
> @leogr can we get this in a release even if beta?

This will be shipped with Falco 0.41 (due by May 26th).

PR #275, opened 2025-03-01
rules: add support for k3s to containerd_activities macro

kyrofa:
**What type of PR is this?**

/kind feature

/area rules

/area maturity-stable

K3s is a stripped down version of Kubernetes that bundles dependencies within it, including containerd. It puts containerd files (sockets, tmpmounts, snapshotter overlayfs, etc.) in namespaced, non-standard locations in an attempt to not interfere with a system-wide containerd installation. As a result, the "Clear Log Activities" rule triggers warnings for the bundled containerd. Fix that by including K3s' non-standard paths in the containerd_activities macro.


darryk10 (approved):

kyrofa:
Thanks folks! How do I actually track this to a release so I can know when I can drop my customized rules?

PR #280, opened 2025-04-02
update(rules): require container plugin.

FedeDP:
**What type of PR is this?**

/kind cleanup

/area rules

Requires `container` plugin in Falco rules.
Also, drop `%container.info` from default ruleset since it now resolves to empty string anyway.

I think that we will need to tag a new maj version for all the rules since this is a breaking change.
Leaving it `wip` for now.
cc @leogr 

Fixes #


rules/falco-deprecated_rules.yaml
```diff
@@ -27,6 +27,10 @@
 # However the Falco rules file does not use them by default.
 - required_engine_version: 0.31.0
 
+- required_plugin_versions:
+    - name: container
+      version: 0.2.0
```
FedeDP, line 32: Please note that 0.2.0 is still not released. This will be the final release that will go into Falco 0.41.0.

FedeDP:
CI is failing because:
```
Error: Plugin 'container' version '0.1.0' is not compatible with required plugin version '0.2.0'\n"
```
that's ok because current Falco master ships with 0.1.0.

leogr:
TODO: we also should remove all references to `container.info`
I would remove the fake grouping using `(` and `)` at this point. Ping me when working on this.

FedeDP:
> TODO: we also should remove all references to container.info

They are already removed in bb5513e253ded8ed48d141948f6797206a234740

leogr (approved):

PR #295, opened 2025-06-13
fix(rules): fixed `container_started` macro adapting to new container plugin

FedeDP:
**What type of PR is this?**

/kind bug

/area rules

/area maturity-incubating
/area maturity-sandbox

In `container_started` macro, avoid using the meta, internal `container` event to check for spawned containers, since the new container plugin will send that event **before** the first process inside the container has spawned.
Instead, just rely on `vpid==1` and `container.id!=host`.

See https://github.com/falcosecurity/falco/issues/3610

Fixes #

### DIFF
--- a/rules/falco-incubating_rules.yaml
+++ b/rules/falco-incubating_rules.yaml
@@ -296,9 +296,7 @@
 
 - macro: container_started
   condition: >
-    ((evt.type = container or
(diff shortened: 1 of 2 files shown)


FedeDP:
Tested with Falco 0.41.0.

PR #299, opened 2025-06-21
Add Rule: Detect Core Pattern Pipe Abuse for Container Escape (T1611)

vVv-Keys:
This PR adds a new detection rule for identifying attempts to abuse `/proc/sys/kernel/core_pattern` by redirecting core dumps to a pipe. This technique is often used to escalate privileges or escape containers when the sysctl setting is writable and misconfigured.

**What type of PR is this?**

> /kind feature

> /area rules

> /area maturity-incubating

Adds a rule that detects write access to `/proc/sys/kernel/core_pattern` containing a pipe symbol (`|`), which may be used to redirect core dumps to arbitrary commands. This has been used in container breakout scenarios and privilege escalations.

### Rule Summary:
- **Rule name:** `Core Pattern Escape Attempt`
- **Detection:** `open_write` on `/proc/sys/kernel/core_pattern` with a pipe (`|`) in the buffer
- **MITRE ATT&CK:** T1611 – Escape to Host
- **Impact:** High – potential root code execution via crash-triggered payloads
None currently tracked — original contribution.

Fixes #

N/A

This rule aligns with documented techniques seen in tools like [CDK](https://github.com/cdk-team/CDK) and various privilege escalation labs. Let me know if you’d like the rule expanded to cover syscall write directly or enriched with container scope conditions.

### DIFF
--- a/rules/core_pattern_escape_attempt.yml
+++ b/rules/core_pattern_escape_attempt.yml
@@ -11,5 +11,5 @@
   output: >
     Core pattern modification to pipe detected | file=%fd.name user=%user.name user_uid=%user.uid user_loginuid=%user.loginuid process=%proc.name proc_exepath=%proc.exepath parent=%proc.pname command=%proc.cmdline terminal=%proc.tty evt_type=%evt.type
   priority: CRITICAL
-  tags: [process, container, escape, mitre_escape, T1611]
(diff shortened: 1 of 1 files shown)


vVv-Keys (commented):
/approve

vVv-Keys (commented):
rules/core_pattern_escape_attempt.yml

PR #363, opened 2026-04-20
feat(rules): detect container shell spawn for MITRE T1059

JayKnowSo:
Adds detection rule for shell processes spawned inside containers by non-shell parent processes — a common indicator of container escape attempts and command injection exploitation.

MITRE ATT&CK: T1059 - Command and Scripting Interpreter
Tags: container, shell, mitre_execution

**What type of PR is this?**

> /kind feature

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /area rules

> /area registry

> /area build

> /area documentation

> /area maturity-stable

> /area maturity-incubating

> /area maturity-sandbox

> /area maturity-deprecated

Fixes #

### DIFF
--- a/rules/falco-incubating_rules.yaml
+++ b/rules/falco-incubating_rules.yaml
@@ -1007,6 +1007,40 @@
 # when more than one event type is involved because some event will populate
 # the filtercheck and others will always return <NA>. It would be better to use
 # a more generic filter like `fs.path.*`
+# This rule is focused on detecting shell history deletion, frequently used by unsophisticated adversaries to eliminate evidence.
+# Note that it can also trigger when exiting a Terminal shell, such as with `kubectl 
+# exec`, which may introduce some noise.
+- macro: user_known_security_tool_disable_activities
+  condition: (never_true)
+
+- rule: Defense Tool Disabled or Modified in Container
+  desc: >
+    Detect attempts to disable or modify security tooling inside a running container,
+    including flushing firewall rules via iptables or stopping security daemons such
+    as falco, auditd, or sysdig. Adversaries impair defenses after achieving initial
+    execution to operate undetected before lateral movement.
+    Maps to MITRE ATT&CK T1562.001 (Impair Defenses: Disable or Modify Tools).
+  condition: >
+    spawned_process and container
+    and (
+      (proc.name in (iptables, ip6tables) and
+       ((proc.args startswith "-F" or proc.args contains " -F") or
+        proc.args contains "--flush" or
+        proc.args contains "-X" or proc.args contains "--delete-chain"))
+      or
+      (proc.name = systemctl and proc.args startswith "stop " and
+       (proc.args contains " falco" or proc.args contains " auditd" or
+        proc.args contains " sysdig" or proc.args contains " osquery"))
+      or
+      (proc.name = service and proc.args startswith "stop " and
+       (proc.args contains " falco" or proc.args contains " auditd"))
+    )
+    and not user_known_security_tool_disable_activities
+  output: Security tool disabled or firewall rules cleared in container | proc=%proc.name args=%proc.args evt_type=%evt.type user=%user.name user_uid=%user.uid user_loginuid=%user.loginuid process=%proc.name proc_exepath=%proc.exepath parent=%proc.pname command=%proc.cmdline terminal=%proc.tty container_id=%container.id image=%container.image.repository
+  priority:
+    WARNING
+  tags: [maturity_incubating, container, process, mitre_defense_evasion, T1562.001]
+
 - rule: Delete or rename shell history
   desc: >
     Detect shell history deletion, frequently used by unsophisticated adversaries to eliminate evidence.
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -1263,3 +1263,22 @@
   output: Fileless execution via memfd_create | container_start_ts=%container.start_ts proc_cwd=%proc.cwd evt_res=%evt.res proc_sname=%proc.sname gparent=%proc.aname[2] evt_type=%evt.type user=%user.name user_uid=%user.uid user_loginuid=%user.loginuid process=%proc.name proc_exepath=%proc.exepath parent=%proc.pname command=%proc.cmdline terminal=%proc.tty exe_flags=%evt.arg.flags
   priority: CRITICAL
   tags: [maturity_stable, host, container, process, mitre_defense_evasion, T1620]
+
+# MITRE ATT&CK T1059 — Container Escape via Command Execution
+# Detects shell spawned inside container by non-shell parent process
+- rule: Detect Shell Spawn in Container (T1059)
+  desc: >
+    A shell was spawned inside a container by a process that is not
+    itself a shell. This is a common indicator of container escape
+    attempts or command injection exploitation.
+  condition: >
+    spawned_process and container and
+    shell_procs and not proc.pname in (shell_binaries) and
+    not container.image.repository in (trusted_images)
+  output: >
+    Shell spawned in container by non-shell parent
+    (user=%user.name container=%container.name
+    image=%container.image.repository
+    parent=%proc.pname shell=%proc.name cmdline=%proc.cmdline)
+  priority: WARNING
+  tags: [container, shell, mitre_execution, T1059]
--- a/rules/falco_rules.yaml
+++ b/rules/falco_rules.yaml
@@ -1278,7 +1278,7 @@
   output: >
     Shell spawned in container by non-shell parent
     (user=%user.name container=%container.name
-    image=%container.image.repository
-    parent=%proc.pname shell=%proc.name cmdline=%proc.cmdline)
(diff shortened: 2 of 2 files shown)


leogr:
Hey @JayKnowSo 

Can you sign-off your commit, please? :pray:

JayKnowSo:
All four addressed.
Cleaned the output fields — dropped image= and args=, added exe_flags per the spawned_process convention.
Anchored the stop args with startswith and put leading spaces on the daemon names. Caught the same pattern on the iptables -F flag — first position and mid-args both covered now.
Dropped network from the tags. Rule never touches a socket — shouldn't have been there.
Also squared away the DCO signoff on both commits — apologies for missing that.
Pushed. Thanks for the welcome and the patience.

JayKnowSo:
CI failures appear to be GitHub Actions infrastructure issues — internal server errors on action archive downloads and a 403 on the repo access. Not related to the rule changes. Happy to re-run once the CI stabilizes.

JayKnowSo:
No yamllint violations in the code added by this PR. The 53 errors in
`falco_rules.yaml` and 67 in `falco-incubating_rules.yaml` are pre-existing
and not introduced by these changes.

Happy to open a separate chore PR to fix the line-length violations across
the affected rule files if that would help unblock the merge pipeline.

rules/falco-incubating_rules.yaml
```diff
@@ -1007,6 +1007,40 @@
...
+      or
+      (proc.name = service and proc.args startswith "stop " and
+       (proc.args contains " falco" or proc.args contains " auditd"))
+    )
+    and not user_known_security_tool_disable_activities
+  output: Security tool disabled or firewall rules cleared in container | proc=%proc.name args=%proc.args evt_type=%evt.type user=%user.name user_uid=%user.uid user_loginuid=%user.loginuid process=%proc.name proc_exepath=%proc.exepath parent=%proc.pname command=%proc.cmdline terminal=%proc.tty container_id=%container.id image=%container.image.repository
+  priority:
+    WARNING
+  tags: [maturity_incubating, container, process, mitre_defense_evasion, T1562.001]
+
```
leogr, line 1043: May you drop this whole block from this PR?

It is the `T1562.001` macro and rule from #367, plus three comment lines above it (1010-1012) that duplicate the `desc` of `Delete or rename shell history` below. I believe the fixes from my #367 review landed on this branch by mistake. Let's keep them in #367 and this PR focused on `T1059`.

When porting them over, please check my latest comments on #367: `-X` is still unanchored here, and `service falco stop` is not matched by `proc.args startswith "stop "` (the SysV syntax puts the verb last).

Btw, the only two new yamllint violations of this PR (trailing space at line 1011 and the long `output` at line 1039) are both in this block, so they go away with it.

rules/falco_rules.yaml
```diff
@@ -1263,3 +1263,22 @@
...
   priority: CRITICAL
   tags: [maturity_stable, host, container, process, mitre_defense_evasion, T1620]
+
+# MITRE ATT&CK T1059 — Container Escape via Command Execution
+# Detects shell spawned inside container by non-shell parent process
+- rule: Detect Shell Spawn in Container (T1059)
+  desc: >
+    A shell was spawned inside a container by a process that is not
+    itself a shell. This is a common indicator of container escape
+    attempts or command injection exploitation.
```
leogr, line 1273: Nit. May you drop `(T1059)` from the rule name? No other rule embeds the technique ID in its name, since the tag already carries it.

Also, T1059 is "Command and Scripting Interpreter", not container escape (we use `T1611` for that). I would rephrase the comment and the `desc` accordingly: this detects a shell spawned by a non-shell parent in a container (i.e., a possible RCE or command injection), not an escape attempt.

rules/falco_rules.yaml
```diff
@@ -1263,3 +1263,22 @@
...
+  condition: >
+    spawned_process and container and
+    shell_procs and not proc.pname in (shell_binaries) and
+    not container.image.repository in (trusted_images)
```
leogr, line 1277: I have a doubt about the scope of this condition.

This is `Run shell untrusted` (see above in this file) restricted to containers, but without the `protected_shell_spawner` scoping and without any of its exclusions. So it fires for every shell spawned by any non-shell parent, in every container. For instance:
- any container whose entrypoint is a shell (`sh -c ...`, `bash entrypoint.sh`), since the parent is `runc`, `containerd-shim`, etc. (see the `container_entrypoint` macro)
- any shell started via `kubectl exec` or `docker exec`, interactive or not (`Terminal shell in container` limits this case to `proc.tty != 0` on purpose)
- cron, supervisors, and all the other legit parents listed in the `Run shell untrusted` exclusions

Moreover, `proc.pname exists` is missing. When the parent is unknown to Falco (e.g., dropped events, or the short-lived `runc` parent race described in the `container_entrypoint` comment), `proc.pname` has no value, so `proc.pname in (shell_binaries)` is false, the `not ...` clause is true, and the rule fires anyway. That is why `Run shell untrusted` has that check.

Lastly, `trusted_images` is a legacy list kept only for backward compatibility (see the note above its definition), and it is about privileged containers and sensitive mounts, not shells. A dedicated `user_known_*` macro (or list) would be the usual tuning knob here.

So, what does this rule catch that `Terminal shell in container` and `Run shell untrusted` do not? If the goal is non-interactive shells spawned by application processes in containers, I would narrow it down (e.g., `proc.tty = 0` plus a parent allowlist) and document the tuning in the `desc`. wdyt? :thinking:

rules/falco_rules.yaml
```diff
@@ -1263,3 +1263,22 @@
...
+  output: >
+    Shell spawned in container by non-shell parent
+    (user=%user.name container=%container.name
+    parent=%proc.pname shell=%proc.name cmdline=%proc.cmdline
+    exe_flags=%evt.arg.flags)
```
leogr, line 1282:
```suggestion
  output: Shell spawned in container by non-shell parent | evt_type=%evt.type user=%user.name user_uid=%user.uid user_loginuid=%user.loginuid process=%proc.name proc_exepath=%proc.exepath parent=%proc.pname command=%proc.cmdline terminal=%proc.tty exe_flags=%evt.arg.flags
```

A few alignments with the style guide (https://falco.org/docs/rules/style-guide/#output-fields) and with the other rules:
- single line, sentence first, then `|` and the `key=value` fields (all the other 25 rules in this file follow this format, none uses a folded scalar)
- the standard set of process and user fields (`evt_type`, `user_uid`, `user_loginuid`, `proc_exepath`, and `terminal` were missing)
- no container fields in upstream outputs, since adopters add them via `append_output` (or `-pc`)

rules/falco_rules.yaml
```diff
@@ -1263,3 +1263,22 @@
...
+  priority: WARNING
+  tags: [container, shell, mitre_execution, T1059]
```
leogr, line 1284:
```suggestion
  tags: [maturity_sandbox, container, shell, mitre_execution, T1059]
```

As per our maturity framework (https://github.com/falcosecurity/rules/blob/main/CONTRIBUTING.md#overall-guidelines), the first tag must be the maturity level, and a new rule cannot start as `maturity_stable`. Since only stable rules live in `falco_rules.yaml`, may you move the whole rule to `falco-sandbox_rules.yaml`?

N.B. `shell_procs` is not defined in the sandbox file. The other sandbox rules just use `proc.name in (shell_binaries)`, see `Web Server Spawned Shell`.

leogr (commented):
Hey @JayKnowSo

Thanks for this PR, and apologies for the long silence on our side 🙏

Before getting into the rule itself, two structural things:

 - The `T1562.001` macro and rule added to `falco-incubating_rules.yaml` are the subject of your other PR (#367). I believe the fixes from my review over there landed on this branch by mistake (the
     "All four addressed" comment ended up here too). May you drop them from this PR and carry them over to #367? Otherwise, whichever PR lands first will conflict with the other one.
 - A new rule cannot start at `maturity_stable`, and only stable rules live in `falco_rules.yaml` (see [https://github.com/falcosecurity/rules/blob/main/CONTRIBUTING.md#overall-guidelines). ](https://github.com/falcosecurity/rules/blob/main/CONTRIBUTING.md#overall-guidelines).)May you move the rule to `falco-sandbox_rules.yaml` with `maturity_sandbox` as its first tag? Also, please uncomment the `/kind feature`, `/area rules`, and `/area maturity-sandbox` lines in the PR description, so the bot can label the PR.

Regarding the rule itself, I have a doubt about its scope. As written, it is `Run shell untrusted` restricted to containers, but without the `protected_shell_spawner` scoping and without any exclusion, so it fires for every shell spawned by a non-shell parent in every container (container entrypoints and `kubectl exec` included). It would help to clarify what it catches that `Terminal shell in container` and `Run shell untrusted` do not. See inline for details.

Also, a few minor alignments with the style guide, see inline :point_down:

N.B. The CI failures are GitHub infrastructure hiccups from May (your assessment was correct), and Yamllint is already red on `main`, so nothing to do on that front. Your next push will re-trigger everything.

PR #364, opened 2026-04-29
docs: add STYLE_GUIDE.md for rules doc comments (closes #150)

### LINKED ISSUE #150: Falco Rules Doc Comments
**Motivation**

The issues I found with comments are due to practical reasons. In particular:

- There should be no empty lines between the comments and the referred item
  - This will avoid ambiguities. 
  - I guess I mostly fixed that with this PR.
- There's no standard item's comment structure
  - I would love to define a simple _Rule Doc_ convention (i.e., something similar to [Go Doc Comments](https://go.dev/doc/comment), but simpler)
- There's no clear way to distinguish sections from item's comment 

**Feature**

Implement a simple _Rules Doc Comments_ guideline and add it to the rules style guide.

**Alternatives**

Just rely on an existing convention, if any, or do nothing. However they don't look like compelling alternatives.

**Additional context**

See https://github.com/falcosecurity/rules/pull/149#issuecomment-1704317483

koiakoia:
Closes #150.

As discussed in the issue, this introduces a `STYLE_GUIDE.md` covering rule documentation conventions:

- Doc comments vs section comments (blank-line rule per @leogr's February 2026 proposal)
- First-sentence-as-summary convention for doc comments
- Override examples in tuning macros (`user_*`, `allowed_*`) — concrete `append: true` snippets
- Source attribution for list-derived data (commands, vendor URLs)
- "Explain or delete" convention for commented-out items

Each section pairs a GOOD example with a BAD example so the convention is unambiguous on review. The examples are drawn from actual items in `falco_rules.yaml` rather than invented syntax.

Also:
- Added a short pointer in `CONTRIBUTING.md` near the top (just after the license note) so contributors find the style guide without hunting for it
- Added `STYLE_GUIDE.md` to `mkdocs.yml` so it renders in the docs site

These conventions emerged from running ~77 custom rules in production on OKD, so I expect the maintainers will have refinements based on a wider rule corpus. Happy to iterate on structure, examples, or scope during review.

/cc @leogr @LucaGuerra @Issif


koiakoia:
Missed DCO on the initial push — force-pushed `af134ef` with proper `Signed-off-by` trailer. Label flipped green. Will run `pre-commit install` first next time.

leogr:
cc @falcosecurity/core-maintainers

STYLE_GUIDE.md
```diff
@@ -0,0 +1,388 @@
...
+
+Tuning macros are placeholders — their default condition is `(never_true)` or
+`(always_true)`, and operators are expected to override them locally to suppress
+false positives. Because these macros have no `desc` field, their doc comment is
+the *only* documentation an operator will read before deciding how to override.
+
+**Required in every tuning macro doc comment:**
+
+1. What the macro does (one sentence).
+2. How to override it — include a concrete `append: true` snippet.
```
ekoops: Actually, `append: true` has been deprecated in favor of `override: {replace|append}`. Could you please adjust this section? `append: true` is also used in other sections, so it would be great if you can fix it also there.

docs/STYLE_GUIDE.md
ekoops, line 1: Shouldn't this be inside docs/ for mkdocs to find it? 🤔

STYLE_GUIDE.md
```diff
@@ -0,0 +1,388 @@
...
+
+```yaml
+# Allowlist of hosts permitted to initiate SSH connections in the monitored rule.
+# Default is never_true; the rule uses double negation (and not allowed_ssh_hosts),
+# so this macro effectively allows everything when unset.
+#
+# To restrict SSH to known management hosts:
+#
+#   - macro: allowed_ssh_hosts
+#     condition: (evt.hostname in (bastion.example.com, mgmt.example.com))
```
ekoops: Mmm this example uses `evt.hostname`, that is the name of the host running Falco, not the destination host. This means that this macro doesn't provide the described behaviour.

ekoops (changes requested):

koiakoia:
@ekoops thank you for the careful review — both points well-taken. Pushed two commits:

**`be40a7e` — review feedback: override syntax + correct SSH allowlist example**

- Replaced `append: true` with `override: append` throughout (Section 3 examples, Section 5 commented-out template, and the Quick Reference table). Added a note that `override: replace` is the swap-default counterpart so operators see both forms.
- Rewrote the `allowed_*` style example. You are correct that `evt.hostname` is the Falco runner's own hostname, so the original macro could never restrict SSH peers as described. Renamed the macro to `allowed_ssh_initiators` and changed the condition to `proc.name in (ansible-playbook, scp, rsync)` — a tuning shape that matches what existing Falco rules actually use. Also called out `fd.rip` in the common-override-fields note for the peer-IP allowlist variant in case operators reach for that.

**`79f6c33` — Move STYLE_GUIDE.md into docs/ for mkdocs discovery**

Good catch — mkdocs's default `docs_dir` is `docs/`, so the top-level file would not have been picked up. The existing `mkdocs.yml` nav entry `Style Guide: STYLE_GUIDE.md` resolves to `docs/STYLE_GUIDE.md` under that default, so the move suffices without rewriting the nav. Added a small comment in `mkdocs.yml` so the path resolution is obvious to future readers.

Happy to make further changes if you spot anything else.

CONTRIBUTING.md
```diff
@@ -6,6 +6,10 @@ This repository includes a dedicated guide for contributing rules, outlining the
 
 All rules must be licensed under the [Apache 2.0 License](./LICENSE).
 
+For comment and documentation conventions within rule files (doc comments,
+section comments, tuning-macro override examples, and list provenance), see
+**[STYLE_GUIDE.md](STYLE_GUIDE.md)**.
```
leogr, line 11:
```suggestion
**[STYLE_GUIDE.md](docs/STYLE_GUIDE.md)**.
```

Since the file lives under `docs/` now, this link needs the prefix, otherwise it 404s from `CONTRIBUTING.md` 🙏

docs/STYLE_GUIDE.md
```diff
@@ -0,0 +1,395 @@
...
+
+Tuning macros are placeholders — their default condition is `(never_true)` or
+`(always_true)`, and operators are expected to override them locally to suppress
+false positives. Because these macros have no `desc` field, their doc comment is
+the *only* documentation an operator will read before deciding how to override.
+
+**Required in every tuning macro doc comment:**
+
+1. What the macro does (one sentence).
+2. How to override it — include a concrete `override: append` snippet.
```
leogr, line 140:
```suggestion
2. How to override it — include a concrete `override:` block snippet (e.g. `condition: append`).
```

Same syntax point as the inline comments below 🙏

docs/STYLE_GUIDE.md
```diff
@@ -0,0 +1,395 @@
...
+# Suppresses "Read sensitive file untrusted" for known-legitimate readers.
+# Default is never_true (no suppression). To allow specific processes:
+#
+#   - macro: user_known_read_sensitive_files_activities
+#     condition: (proc.name in (my_backup_agent, my_audit_tool))
+#     override: append
+#
+# Use `override: replace` to swap the default condition entirely instead of
+# extending it. Common override fields: proc.name, proc.exepath,
+# container.image.repository.
```
leogr, line 156:
```suggestion
#     override:
#       condition: append
#
# Use `condition: replace` (instead of `append`) to swap the default condition
# entirely. Common override fields: proc.name, proc.exepath,
# container.image.repository.
```

`override` is a block in the canonical syntax (the value selects per-field append/replace). For a macro it is `condition: append`; for a list, `items: append`. See 👉 https://falco.org/docs/concepts/rules/overriding/

I verified with `falco -V` on 0.43.0: the single-value form (`override: append`) loads but produces schema warnings, so I would rather we teach the right shape from day one 🙏

docs/STYLE_GUIDE.md
```diff
@@ -0,0 +1,395 @@
...
+# To restrict SSH initiation to known automation tooling:
+#
+#   - macro: allowed_ssh_initiators
+#     condition: (proc.name in (ansible-playbook, scp, rsync))
+#     override: append
+#
+# Use `override: replace` to swap the default condition entirely instead of
+# extending it. Common override fields: proc.name, proc.exepath, fd.rip
+# (for peer-IP allowlists). See also: macro never_true for the placeholder
+# pattern.
```
leogr, line 192:
```suggestion
#     override:
#       condition: append
#
# Use `condition: replace` (instead of `append`) to swap the default condition
# entirely. Common override fields: proc.name, proc.exepath, fd.rip
# (for peer-IP allowlists). See also: macro never_true for the placeholder
# pattern.
```

Same here 🙏

docs/STYLE_GUIDE.md
```diff
@@ -0,0 +1,395 @@
...
+
+```yaml
+# Optional: treat any node process in a container as a protected shell spawner.
+# Disabled by default because node is also widely used as a build-pipeline tool
+# where spawning shells is expected. Enable by overriding this macro to remove
+# the never_true guard:
+#
+#   - macro: possibly_node_in_container
+#     condition: (proc.pname=node and proc.aname[3]=docker-containe)
+#     override: replace   # swap the default never_true guard entirely
```
leogr, line 304:
```suggestion
#     override:
#       condition: replace   # swap the default never_true guard entirely
```

Same here 🙏

docs/STYLE_GUIDE.md
```diff
@@ -0,0 +1,395 @@
...
+
+---
+
+## Quick Reference
+
+| Situation | Convention |
+|-----------|------------|
+| Comment immediately precedes an item (no blank line) | Doc comment — documents that item |
+| Comment has a blank line before it AND below it | Section comment — labels a group |
+| Tuning macro (`user_*`, `allowed_*`) | Doc comment must include override snippet with `override: append` (or `override: replace`) |
```
leogr, line 378:
```suggestion
| Tuning macro (`user_*`, `allowed_*`) | Doc comment must include override snippet using an `override:` block (e.g. `condition: append` or `condition: replace`) |
```

Same syntax fix as above 🙏

docs/STYLE_GUIDE.md
```diff
@@ -0,0 +1,395 @@
...
+| First sentence of any doc comment | Summarize purpose; avoid "This macro…" preamble |
+
+---
+
+## Relationship to the Existing Style Guide
+
+This guide covers **comment conventions only**. For rule expression style,
+output format, tag requirements, and maturity-level criteria, see the
+[Falco Rules Style Guide](https://falco.org/docs/rules/style-guide/) and
+[CONTRIBUTING.md](CONTRIBUTING.md).
```
leogr, line 391:
```suggestion
[CONTRIBUTING.md](../CONTRIBUTING.md).
```

`CONTRIBUTING.md` is at the repo root, so from here it needs `../`. Otherwise the relative link breaks 🙏

PR #383, opened 2026-09-02
fix(.github): validate fork PR numbers in create-comment workflow

ManuelFCastillo:
Same fix as falcosecurity/plugins#1509, for the identical workflow in this repository. Reported in falcosecurity/plugins#1505, where @leogr noted this repo needs it too.

### Problem

The validation in `create-comment.yaml` calls `listPullRequestsAssociatedWithCommit` against `context.repo`, which in a `workflow_run` event is always the base repository. For a pull request opened from a fork the head commit lives in the contributor's fork, so the base repo lookup returns an empty list, `valid_pr_numbers` is empty, and the check can never pass.

Branches pushed directly to this repository are unaffected, which is why it goes unnoticed.

### Change

Ports the approach already in use in falcosecurity/libs (falcosecurity/libs@0501075).

- For same-repo pull requests, `workflow_run.pull_requests` is populated and is used directly.
- For fork pull requests it is always empty, so fall back to `pulls.list` on `context.repo` filtered by `head: "<head_repository.owner.login>:<head_branch>"`.

Both values come from the `workflow_run` payload and are set by GitHub rather than by the fork.

### Why this preserves the original guarantee

`pulls.list` is called on the base repository, so every candidate number is a pull request of this repository by construction. No additional `base.repo` or `head.sha` filtering is needed.

The resulting block is byte-identical to the libs implementation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/create-comment.yaml
+++ b/.github/workflows/create-comment.yaml
@@ -48,15 +48,30 @@ jobs:
 
             // Validate artifact PR number against the workflow_run context
             // to prevent artifact poisoning (cross-issue comment injection).
-            const head_sha = context.payload.workflow_run.head_sha;
-            const {data: associated_prs} = await github.rest.repos.listPullRequestsAssociatedWithCommit({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              commit_sha: head_sha,
-            });
-            const valid_pr_numbers = associated_prs.map(pr => pr.number);
+            //
+            // For same-repo PRs, workflow_run.pull_requests is populated.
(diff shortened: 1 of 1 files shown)


leogr:
Same as https://github.com/falcosecurity/plugins/pull/1509, identical to the libs implementation. Thank you 🙏

N.B. The `Yamllint` failure is unrelated to this PR (it only checks `rules/*.yaml`, and it's currently failing on every PR).

/lgtm
/approve

leogr (approved):

